PR #6, opened 2019-10-09
WIP: Invader AI

parasyte:
- Adds debug mode for visualizing bounding boxes
- Adds rectangle and line drawing (for debug mode)
- Invaders move as a close approximation to the original game
- TODO: Demonstrates that the blit function needs to ignore black pixels (or "transparency")
- TODO: The invader movement code is really bad


parasyte:
There's a bug revealed by the player movement scaling. Seems like winit gets into a state where it is grossly unsynchronized with the display, and it causes lots of janky player movement with this algorithm. Restarting the game fixes it. 😐

PR #37, opened 2019-11-06
Add a conways game of life example

thomcc:
I wanted to write one of these since someone mentioned it in discord, and your lib made it easy to do. Some of this code is snarfed from the other example. It has a bug that I mentioned in discord, but once you resize, the position computation for the mouse is wrong. I don't see a way to correct it in the public API.


examples/conway/main.rs
```diff
@@ -0,0 +1,394 @@
...
+            }
+            if !paused || input.key_pressed(VirtualKeyCode::Space) {
+                life.update();
+            }
+            window.request_redraw();
+        }
+    });
+}
+
+// COPYPASTE: ideally this could be shared.
```
parasyte: Agreed! I considered something like the "framework" pattern used for some other crates examples. It would be a nice addition to DRY it up.

examples/conway/main.rs
```diff
@@ -0,0 +1,394 @@
...
+    fn grid_idx<I: std::convert::TryInto<usize>>(&self, x: I, y: I) -> Option<usize> {
+        if let (Ok(x), Ok(y)) = (x.try_into(), y.try_into()) {
+            Some(x + y * self.width)
+        } else {
+            None
+        }
+    }
+}
+
+fn bresenham(x0: isize, y0: isize, x1: isize, y1: isize, mut cb: impl FnMut(isize, isize) -> bool) {
```
parasyte: I honestly considered doing this, too. But took a shortcut when I found the [`line_drawing`](https://docs.rs/line_drawing/0.8.0/line_drawing/struct.Bresenham.html) crate, which is already included as a dev-dependency. Maybe worth just reusing that?

examples/conway/main.rs
thomcc: Yeah, that's an option, but OTOH it's nice to be able to have relatively simple example code that doesn't require figuring out a bunch of parts in order to get a window up and things rolling. I don't really care tho, your call. I think that should be handled separately from this PR though.

examples/conway/main.rs
```diff
@@ -0,0 +1,394 @@
...
+                .map(|(mx, my)| {
+                    let (dx, dy) = input.mouse_diff();
+                    let prev_x = mx - dx;
+                    let prev_y = my - dy;
+                    let dpx = hidpi_factor as f32;
+                    let (w, h) = (p_width as f32 / dpx, p_height as f32 / dpx);
+                    let mx_i = ((mx / w) * (SCREEN_WIDTH as f32)).round() as isize;
+                    let my_i = ((my / h) * (SCREEN_HEIGHT as f32)).round() as isize;
+                    let px_i = ((prev_x / w) * (SCREEN_WIDTH as f32)).round() as isize;
+                    let py_i = ((prev_y / h) * (SCREEN_HEIGHT as f32)).round() as isize;
```
parasyte: I see what you mean. I would love to have this kind of logic baked into the public API in some form. Admittedly it's just a feature I didn't think of at the time. But seems so obvious now.

examples/conway/main.rs
```diff
@@ -0,0 +1,394 @@
...
+            }
+            // Adjust high DPI factor
+            if let Some(factor) = input.hidpi_changed() {
+                hidpi_factor = factor;
+            }
+            // Resize the window
+            if let Some(size) = input.window_resized() {
+                let size = size.to_physical(hidpi_factor);
+                let width = size.width.round() as u32;
+                let height = size.height.round() as u32;
```
parasyte: nitpick, but you could just reuse the `p_width` and `p_height` captured variables, no?

examples/conway/main.rs
parasyte: Another example that I've been intending to add is a dead simple "hello" example that doesn't do anything fancy with events or anything. It would just draw something - a rectangle or "hi", something that doesn't require much code. Perhaps that could serve as the kind of example you're talking about. At least then it wouldn't feel so dirty adding another complex set of 
abstractions to example code?

Perfectly reasonable to make those adjustments another time.

examples/conway/main.rs
thomcc: Nice catch.

examples/conway/main.rs
thomcc: Sure. I didn't notice the dep existed, and wasn't about to add a new one for a short function.

PR #41, opened 2019-11-07
Hypothesis: Removing the Travis cache should make CI builds much faster

parasyte:


parasyte:
I believe that's a confirmation

Issue #40, opened 2019-11-07
Safety dance!


parasyte:
See [Rust Safety Dance](https://github.com/rust-secure-code/safety-dance/) - it's cute!

There is one bit of `unsafe` code, and that's the `Deref` impl for the `Spirv` type: https://github.com/parasyte/pixels/blob/e0225c145d2972b8803e64088fad9de0951caf02/src/macros.rs#L34-L36

Making `pixels` 100% safe Rust is kind of a hilarious challenge in an ironic way, considering it sits on top of a mountain of `unsafe`.

Huanzo:
#43 Here's my take on this :D Its 100% safe now but it comes at a slight startup performance penalthy. We should find out if it is noticable/matters much.

PR #43, opened 2019-11-07
Remove Unsafe in macro.rs

Huanzo:
This is my take on removing the unsafe block in macro.rs.

This will definitely be slower than the unsafe version but I don't think that it matters too much as it will not be called very often.

I could not get it work with the Deref impl so I changed that to a normal  function.


src/macros.rs
```diff
@@ -27,13 +27,18 @@ macro_rules! include_spv {
...
-            std::slice::from_raw_parts(self.0.as_ptr() as *const u32, self.0.len() / 4)
+impl Spirv {
+    pub fn deref(&self) -> Vec<u32> {
+        let mut out = Vec::with_capacity(self.0.len() / 4);
+        for i in 0..self.0.len() / 4 {
+            let mut tmp: u32 = 0;
+            tmp += self.0[4 * i] as u32;
+            tmp += (self.0[4 * i + 1] as u32) << 8;
+            tmp += (self.0[4 * i + 2] as u32) << 16;
+            tmp += (self.0[4 * i + 3] as u32) << 24;
```
bjorn3: This will have different behaviour compared to the original on big endian systems.

Huanzo:
Even the original did not account for endienness mismatch. So I think this
should not break on any system that it previously worked on.

bjorn3 <notifications@github.com> schrieb am Do., 7. Nov. 2019, 23:53:

src/macros.rs
```diff
@@ -27,13 +27,18 @@ macro_rules! include_spv {
...
 
-impl std::ops::Deref for Spirv {
-    type Target = [u32];
-    fn deref(&self) -> &[u32] {
-        #[allow(clippy::cast_ptr_alignment)]
-        unsafe {
-            std::slice::from_raw_parts(self.0.as_ptr() as *const u32, self.0.len() / 4)
+impl Spirv {
+    pub fn deref(&self) -> Vec<u32> {
+        let mut out = Vec::with_capacity(self.0.len() / 4);
```
parasyte: If we are going to allocate a vector anyway, there's always [`wgpu::read_spirv`](https://docs.rs/wgpu/0.4.0/wgpu/fn.read_spirv.html). And it will handle the endianness correctly, too.

Consequently, this is what was used before the macro: https://github.com/parasyte/pixels/blob/9f1afdd3b324ccbe68556bca36253449c2185cb7/src/lib.rs#L113-L120

We also don't need to use `Deref` if we are removing the `slice::from_raw_parts`.

parasyte:
In addition, we can also add `#![forbid(unsafe_code)]` to lib.rs with the last bit of unsafe removed.

Huanzo:
Ah ok I did not see that.

Why did you choose to go the unsafe route then? Performance?

parasyte:
No, it was contributed by Ralith in their PR for ash: https://github.com/MaikKlein/ash/pull/245

Huanzo:
Ok i now have used `u32::from_ne_bytes()` to repect the target endianness. 

Do we want/need to throw that output Vec into `wgpu::read_spirv` or can we assume that the resulting bytecode is valid? Thing is that `wgpu::read_spirv` takes something that is `Read` and `Seek` which limits our options quite a bit. So will this suffice for now?

parasyte:
`wgpu::read_spirv` handles endianness correctly regardless of endianness of the SPIRV _and_ endianness of the host. See https://github.com/gfx-rs/gfx/blob/ca53a530843c8b3f76928b42753113f1a785b101/src/hal/src/pso/mod.rs#L299-L304

The `Read` and `Seek` trait bounds aren't a problem, since [`std::io::Cursor`](https://doc.rust-lang.org/stable/std/io/struct.Cursor.html) takes care of that.

src/macros.rs
```diff
@@ -7,41 +7,12 @@
...
-///
-/// Does not account for endianness mismatches between the SPIR-V file and the target. See
-/// [`wgpu::read_spirv`] for a more general solution.
 #[macro_export]
 macro_rules! include_spv {
-    ($path:expr) => {
-        &$crate::Align4(*include_bytes!($path)) as &$crate::Spirv
-    };
+    ($path:expr) => {{
+        let arr = include_bytes!($path).to_vec();
```
bjorn3: I think it will work without `.to_vec()`. You may need to use `&*arr` below.

Huanzo: I didn"t get it to work because `std::io::Cursor` needs T to implement `AsRef<[u8]>` and this trait is only implemented for arrays up to a size of 32. Maybe with const generics this will be possible but they havent landed yet.

bjorn3: By doing `&*arr` or `&arr[...]`, the array will turn inti a slice, which implements `AsRef<[u8]>`.

Huanzo: this still wont work because `AsRef<[u8]>` is only implemented for arrays of sizes up to 32.

I'm getting this error when i do what you suggested:
```
error[E0277]: arrays only have std trait implementations for lengths 0..=32
   --> src/macros.rs:15:26
    |
15  |         wgpu::read_spirv(std::io::Cursor::new(&*arr))
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::array::LengthAtMost32` is not implemented for `[u8; 1484]`
```

Huanzo: ah ok now I see what you mean. it works with `&arr[..]`. I thought it was the same as `&*arr`

parasyte: Since the slice is getting copied into Vec<T> anyway, I don’t see a reason not to use Cursor<Vec<u8>>, which satisfies all of the requirements for wgpu::read_spirv

I know it will possibly do more copies and use more heap. But my primary motivation for the macro is to avoid reading from FS at runtime.

Huanzo: Yeah but why take that unnecessary allocation if you can just avoid it that easily.
My latest commit addresses this and also merged changes from master.

parasyte: The most recent patch looks great. Can you also move the borrow inside the macro? Just do we don't have to write `&include_spv!(...)`

Then I think this will be perfect.

parasyte:
Closes #40

Issue #47, opened 2019-11-10
SDL2 support on macOS


parasyte:
There are many crates for interfacing with SDL2. There is only one that also uses `raw_window_handle` and that's [`beryllium`](https://crates.io/crates/beryllium).

When I attempted to port the `minimal-winit` example to SDL2 with `beryllium`, I ran into a showstopper where the `raw_window_handle` impl is not able to provide an `NSView` on macOS, and `wgpu`/`gfx` doesn't try to get the default contentView from the NSWindow. So it sees a NULL pointer and panics with `window does not have a valid contentView`.

I could workaround this issue by making the example depend on `cocoa`, at the expense of adding code that shouldn't be in the example anyway. Another option is building a crate specifically for transforming a `raw_window_handle` into a `raw_window_handle` with an `NSView`. 😐

parasyte:
Fixed by #60 (Thanks @lovesegfault!)

fzzr-:
Unfortunately it isn't fixed for me.
I just tried [minimal-sdl2 example][] and it panicked with `window does not have a valid contentView`.

My env versions:
- OS: macOS 10.15.7 (19H2)
- Xcode: 11.7 (xcode-select: 2373)
- SDL2: 2.0.12 (also 2.0.0)


<details><summary>Full log</summary>
<pre><code>
cd minimal-sdl2
cargo run
...
thread 'main' panicked at 'window does not have a valid contentView', /Users/ak/.cargo/registry/src/github.com-1ecc6299db9ec823/gfx-backend-metal-0.6.5/src/lib.rs:391:13
stack backtrace:
   0:        0x102b03fd4 - std::backtrace_rs::backtrace::libunwind::trace::hd4e4f36eb9f0f866
                               at /rustc/1c389ffeff814726dec325f0f2b0c99107df2673/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:        0x102b03fd4 - std::backtrace_rs::backtrace::trace_unsynchronized::hddd9cf7e61567951
                               at /rustc/1c389ffeff814726dec325f0f2b0c99107df2673/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:        0x102b03fd4 - std::sys_common::backtrace::_print_fmt::hf0df20706c8d0efd
                               at /rustc/1c389ffeff814726dec325f0f2b0c99107df2673/library/std/src/sys_common/backtrace.rs:67:5
[...]

PR #60, opened 2020-04-12
treewide: update dependencies

lovesegfault:
This updates all of `pixels`'s dependencies. Most importantly, it gets us winit 0.22 and wgpu 0.5.0.

This was done in a best-effort way, as I sometimes wasn't entire sure how to get things to work. There's a nasty transmute I want to get rid off, but couldn't figure out how. Figured I'd open the PR to get some ideas.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,8 +58,8 @@ pub struct Pixels {
 }
 
 /// A builder to help create customized pixel buffers.
-pub struct PixelsBuilder {
-    request_adapter_options: wgpu::RequestAdapterOptions,
+pub struct PixelsBuilder<'req> {
+    request_adapter_options: wgpu::RequestAdapterOptions<'req>,
     device_descriptor: wgpu::DeviceDescriptor,
     width: u32,
     height: u32,
@@ -75,6 +75,9 @@ pub enum Error {
     /// No suitable [`wgpu::Adapter`] found
     #[error("No suitable `wgpu::Adapter` found")]
     AdapterNotFound,
+    /// Equivalent to [`wgpu::TimeOut`]
+    #[error("The GPU timed out when attempting to acquire the next texture or if a previous output is still alive.")]
+    Timeout,
 }
 
 impl SurfaceTexture {
@@ -93,10 +96,10 @@ impl SurfaceTexture {
     /// let event_loop = EventLoop::new();
     /// let window = Window::new(&event_loop).unwrap();
     /// let surface = Surface::create(&window);
-    /// let size = window.inner_size().to_physical(window.hidpi_factor());
+    /// let size = window.inner_size();
     ///
-    /// let width = size.width.round() as u32;
-    /// let height = size.height.round() as u32;
+    /// let width = size.width;
+    /// let height = size.height;
     ///
     /// let surface_texture = SurfaceTexture::new(width, height, surface);
     /// # Ok::<(), pixels::Error>(())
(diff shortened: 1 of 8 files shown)


lovesegfault:
Here's the conway example running with the new deps:
image

lovesegfault:
Alright, all warnings are fixed now :)

src/renderers.rs
```diff
@@ -37,45 +37,58 @@ impl Renderer {
...
 
         // Create uniform buffer
         #[rustfmt::skip]
-        let transform: [f32; 16] = [
+        let mut transform: [f32; 16] = [
             1.0, 0.0, 0.0, 0.0,
             0.0, 1.0, 0.0, 0.0,
             0.0, 0.0, 1.0, 0.0,
             0.0, 0.0, 0.0, 1.0,
         ];
```
bjorn3, line 50:
```suggestion
        let transform: [f32; 16] = [
            1.0, 0.0, 0.0, 0.0,
            0.0, 1.0, 0.0, 0.0,
            0.0, 0.0, 1.0, 0.0,
            0.0, 0.0, 0.0, 1.0,
        ];
        let mut transform_bytes: Vec<u8> = Vec::with_capacity(16*4);
        for f in &transform {
            transform_bytes.extend(f.to_bits().to_ne_bytes().iter().copied());
        }
```

This makes it possible to avoid the unsafe code below.

lovesegfault: Done, thanks for the suggestion!

parasyte:
What happened that made `beryllium` stop working with `wgpu`? And if all else fails, can we use the raw bindings (`fermium`) instead?

lovesegfault:
@parasyte if you checkout this branch's first commit you can reproduce the issue:
```
error[E0277]: the trait bound `beryllium::window::gl_window::GlWindow: raw_window_handle::HasRawWindowHandle` is not satisfied
   --> examples/minimal-sdl2/main.rs:32:39
    |
32  |         let surface = Surface::create(&window);
    |                                       ^^^^^^^ the trait `raw_window_handle::HasRawWindowHandle` is not implemented for `beryllium::window::gl_window::GlWindow`
    |
   ::: /home/bemeurer/.cache/cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.5.0/src/lib.rs:596:22
    |
596 |     pub fn create<W: raw_window_handle::HasRawWindowHandle>(window: &W) -> Self {
    |                      ------------------------------------- required by this bound in `wgpu::Surface::create`

error: aborting due to previous error
```

If someone can figure this one out then we don't need to nuke the SDL example :)

lovesegfault:
Oh, wait, I think I figured it out!

src/renderers.rs
```diff
@@ -90,7 +102,7 @@ impl Renderer {
                     binding: 2,
                     resource: wgpu::BindingResource::Buffer {
                         buffer: &uniform_buffer,
-                        range: 0..64,
+                        range: 0..16,
```
parasyte, line 105: Pretty sure this should not be changed!

parasyte:
@lovesegfault It looks like you want to use `create_raw_window` instead of `create_gl_window`

lovesegfault:
@parasyte Correct! Just fixed it :)

src/renderers.rs
lovesegfault: Fixed :)

src/lib.rs
```diff
@@ -183,32 +183,35 @@ impl Pixels {
     /// This executes all render passes in sequence. See [`RenderPass`].
     pub fn render(&mut self) {
         // TODO: Center frame buffer in surface
-        let frame = self.swap_chain.get_next_texture();
+        let frame = self
+            .swap_chain
+            .get_next_texture()
+            .expect("swapchain failed");
```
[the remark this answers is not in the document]
lovesegfault: I made it a panic to avoid changing the API without your okay, I'll change it to an error :)

parasyte:
It looks like [MSRV in CI](https://github.com/lovesegfault/pixels/blob/6de14c3f3ee1f4d3a4377055485156fc4e2747b8/.travis.yml#L3-L4) also needs to be updated to 1.40.0. https://travis-ci.org/github/parasyte/pixels/jobs/674268632

src/lib.rs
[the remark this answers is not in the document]
lovesegfault: Alright, I added error handling

lovesegfault:
@parasyte You may want to consider #61 before this PR, either way I'm fixing Travis :)

parasyte:
It's on my plate to review.

parasyte:
This patch was needed to get `minimal-sdl` working. (Bonus: It works on macOS now! Fixes #47)

lovesegfault:
@parasyte Patch applied :)

parasyte:
This is great! After merging, I'll followup by moving the examples to individual crates to fix the dependency bloat problem wrt. #62

PR #61, opened 2020-04-12
ci: move to github actions

lovesegfault:
This allows us to use all the nice things from GH actions and actions-rs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,84 @@
+name: CI
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '* * * * 0'
+jobs:
+  checks:
+    name: Check
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - stable
+          - beta
+          - 1.40.0
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: sudo apt -y install libsdl2-dev
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          override: true
+      - name: Cargo check
+        uses: actions-rs/cargo@v1
+        with:
+          command: check
+  tests:
+    name: Test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - stable
+          - beta
+          - 1.40.0
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: sudo apt -y install libsdl2-dev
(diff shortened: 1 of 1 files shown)


lovesegfault:
Given the discussion in #60 I'll be adding:
1. MSRV check
2. Multiple platforms

parasyte:
Some other things would be really nice to have here:

1. Should allow failures on `nightly`
2. Perform weekly builds with `on.schedule` to avoid bitrot.
3. `check` should run `cargo fmt --all -- --check` and `cargo doc --workspace --no-deps` to enforce formatting and docs best practices. (I see `fmt` is in `lints` ... might be fine there)
4. `lints` should run `cargo clippy --all --tests -- -D warnings` to deny warnings

lovesegfault:
@parasyte I can't figure out how to allow failures, would you rather have a blocking nightly or no nightly?

lovesegfault:
Alright, I think I got all other requests done :)

parasyte:
Hmmm, good question! I guess I would rather remove nightly, sadly. There is https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error but it looks like you will have to create separate `checks` `tests` jobs for nightly with this.

lovesegfault:
Alright, I fixed the syntax errors :)

lovesegfault:
@parasyte After you merge this, please remember to update the badge in the README, I can't get the correct link until this is merged and the actions window shows up :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,84 @@
...
+      matrix:
+        rust:
+          - stable
+          - beta
+          - 1.40.0
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: sudo apt -y install libudev-dev
```
parasyte, line 21: This will probably have to be updated to match #67:

https://github.com/parasyte/pixels/pull/67/files#diff-354f30a63fb0907d4ad57269548329e3R17

```suggestion
        run: sudo apt -y install libsdl2-dev
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,84 @@
...
```
parasyte, line 45: ditto

.github/workflows/ci.yml
```diff
@@ -0,0 +1,84 @@
...
+        with:
+          command: test
+  lints:
+    name: Lints
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: sudo apt -y install libudev-dev
```
parasyte, line 63: ditto

.github/workflows/ci.yml
lovesegfault: Done :)

PR #62, opened 2020-04-12
cargo: remove features

lovesegfault:
The justification for them existing was that cargo adds in binary-only
deps to mixed bin/lib crates, which inflates dependency trees and
overall just sucks.

While this is true, that doesn't apply to dev-dependencies, which don't
get propagated, but which are available to example code. By removing the
myriad of features and moving those dependencies into `dev-dependencies`
we make developers' lives easier (it's easier to run examples) and don't
actually lose anything.

Moreover, this also allows to test all code with no need for
`--all-features`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,51 +26,19 @@ maintenance = { status = "actively-developed" }
 thiserror = "1.0"
 wgpu = "0.4"
 
-# These are only used by the examples, and enabled with features
-# See: https://github.com/rust-lang/cargo/issues/1982
-beryllium = { version = " 0.1", optional = true }
-byteorder = { version = "1.3", optional = true }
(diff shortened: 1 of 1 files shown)


bjorn3:
This means that **all** dependencies required for specific examples will be build when you try to build **any** of the examples.

lovesegfault:
That's correct, the difference is actually small and the gain in convenience of not having to deal with all the feature flags is huge.

parasyte:
I want to move the examples to their own crates so we don't have dependency bloat. The features were introduced because I didn't want to build e.g. all of the SDL dependencies on the `minimal-winit` example. Moving each example to its own crate solves the dilemma without relying on feature gates.

lovesegfault:
Understood, I'll do that in a separate PR then :)

PR #64, opened 2020-04-13
Update to WGPU 0.5

zicklag:
Hey there, I'm wanting to experiment with Pixels, but I needed OpenGL support because that is all my target computer supports. There is work in progress ( https://github.com/gfx-rs/wgpu/issues/450 ) to land OpenGL support for WGPU 0.5 so I started trying to port Pixels to WGPU 0.5. It was mostly straight forward except when it came the transform buffer data because the buffer now has to be a slice of `u8` instead of a slice of `T` which prevents us from sticking `f32`'s in the transform directly:

https://github.com/katharostech/pixels/blob/d6d6570573cdeb840dd4092d83821eb7a1197788/src/renderers.rs#L46-L58

I ( knowing it would be completely broke ) just replaced the floats with `u8`'s for now to get it to compile, but I'm not exactly how to go about fixing that and could use some help if you get the time.

This is still messy in some spots ( such as the patches in the `Cargo.toml` ) but I'll fix that before I mark it ready to merge.

### DIFF
--- a/src/renderers.rs
+++ b/src/renderers.rs
@@ -37,38 +37,49 @@ impl Renderer {
             mipmap_filter: wgpu::FilterMode::Nearest,
             lod_min_clamp: 0.0,
             lod_max_clamp: 1.0,
-            compare_function: wgpu::CompareFunction::Always,
+            compare: wgpu::CompareFunction::Always,
         });
 
         // Create uniform buffer
+        // TODO: Figure out how to store floats for this transform if necessary
         #[rustfmt::skip]
-        let transform: [f32; 16] = [
-            1.0, 0.0, 0.0, 0.0,
-            0.0, 1.0, 0.0, 0.0,
-            0.0, 0.0, 1.0, 0.0,
-            0.0, 0.0, 0.0, 1.0,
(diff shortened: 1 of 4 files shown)


parasyte:
There another open PR which updates all dependencies, including `wgpu`: https://github.com/parasyte/pixels/pull/60

src/renderers.rs
```diff
@@ -37,45 +37,57 @@ impl Renderer {
...
-        let transform: [f32; 16] = [
-            1.0, 0.0, 0.0, 0.0,
-            0.0, 1.0, 0.0, 0.0,
-            0.0, 0.0, 1.0, 0.0,
-            0.0, 0.0, 0.0, 1.0,
+        let mut transform: [u8; 16] = [
+            1, 0, 0, 0,
+            0, 1, 0, 0,
+            0, 0, 1, 0,
+            0, 0, 0, 1,
```
lovesegfault, line 50: This is incorrect, look for the correct approach in #60

zicklag:
Oh, sweet. Don't need this one then! Thanks @lovesegfault.

PR #67, opened 2020-04-13
Fix dependency bloat

parasyte:
This moves all examples to their own individual crates as a means to avoid dependency bloat without resorting to feature gates.


parasyte:
I could move the `simple-invaders` somewhere else in the source tree, but I'm keen to leave it as its own crate. I have other plans for it later.

lovesegfault:
@parasyte You can leave it as it's own crate in `examples/` I suppose?

parasyte:
@lovesegfault I moved the `simple-invaders` crate. And also got CI back to green.

PR #70, opened 2020-04-13
Fix panic in winit examples when `pixels.render()` returns Error

parasyte:


parasyte:
I'll leave it up to the lib users to handle errors. :D I just don't want the examples to panic.

Issue #75, opened 2020-04-23
Regression: setting wgpu backend is not possible in 0.0.3


parasyte:
The backend is now hardcoded: https://github.com/parasyte/pixels/blob/809d3f6dd9c1c47204c0cc3eac15d814bac731ef/src/lib.rs#L431 (used to be a child of `RequestAdapterOptions`). This needs a new builder method and probably cutting a new release soon.

Aeledfyr:
I believe that this should be closed, as #78 added a builder method for this.  I guess github didn't recognize the second issue mention.

PR #78, opened 2020-05-15
Add builder methods for VSync and setting wgpu backend

### LINKED ISSUE #57: Add builder method for setting VSync
VSync is currently hardcoded on: https://github.com/parasyte/pixels/blob/7701ac2d0e981f07c5cede721c49d4381ff1fa6d/src/lib.rs#L455

It's important to make this optional so that users can decide if they prefer no screen tearing (VSync on) or low latency (VSync off).

Aeledfyr:
* `enable_vsync` sets the VSync mode
* `wgpu_backend` sets the wgpu `BackendBit` struct
* Made `RequestAdapterOptions` use `compatible_surface` by default.  This doesn't do much for most cases, but it may avoid problems for some specific setups and seems to be the correct thing to do.

This fixes #57 and #75.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,6 +46,7 @@ pub struct Pixels {
     queue: Rc<RefCell<wgpu::Queue>>,
     swap_chain: wgpu::SwapChain,
     surface_texture: SurfaceTexture,
+    enable_vsync: bool,
 
     // List of render passes
     renderers: Vec<BoxedRenderPass>,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -301,14 +310,13 @@ impl<'req> PixelsBuilder<'req> {
...
-                compatible_surface: None,
-                power_preference: wgpu::PowerPreference::Default,
-            },
+            request_adapter_options: None,
             device_descriptor: wgpu::DeviceDescriptor::default(),
+            backend: wgpu::BackendBit::PRIMARY,
             width,
             height,
             pixel_aspect_ratio: 1.0,
+            enable_vsync: true,
```
parasyte, line 319: Maybe it's better to just pass around a `wgpu::PresentMode`... I imagine the `Fifo` mode can be useful in some situations!

PR #95, opened 2020-07-18
New rendering API

JMS55:


examples/conway/src/main.rs
```diff
@@ -31,7 +31,9 @@ fn main() -> Result<(), Error> {
         if let Event::RedrawRequested(_) = event {
             life.draw(pixels.get_frame());
             if pixels
-                .render()
+                .render(|encoder, render_target, scaling_renderer| {
+                    scaling_renderer.render(encoder, render_target);
+                })
```
parasyte, line 36: Can we simplify this to `pixels.render(None)`?

src/lib.rs
```diff
@@ -12,21 +12,13 @@
 #![deny(clippy::all)]
 #![forbid(unsafe_code)]
 
-use std::cell::RefCell;
-use std::rc::Rc;
+mod macros;
+mod renderers;
```
parasyte, line 16: Why were these definitions moved?

src/lib.rs
```diff
@@ -42,17 +34,17 @@ pub struct SurfaceTexture {
...
     present_mode: wgpu::PresentMode,
 
-    // List of render passes
-    renderers: Vec<BoxedRenderPass>,
+    // A default renderer to scale the input texture to the screen size
+    scaling_renderer: ScalingRenderer,
 
     // Texture state for the texel upload
-    texture: wgpu::Texture,
+    pixel_texture: wgpu::Texture,
```
parasyte, line 47: Why was this field renamed? It follows that other fields prefixed `texture` should have also adopted the new prefix for consistency. (These are all related.) But I still don't know what benefit we get by renaming this field.

src/lib.rs
```diff
@@ -192,21 +181,42 @@ impl Pixels {
...
-        self.queue.borrow_mut().submit(&[encoder.finish()]);
+        self.scaling_renderer
+            .resize(&mut self.device, &mut encoder, width, height);
+        self.queue.submit(&[encoder.finish()]);
     }
 
-    /// Draw this pixel buffer to the configured [`SurfaceTexture`].
-    ///
-    /// This executes all render passes in sequence. See [`RenderPass`].
+    /// Provides a [`wgpu::CommandEncoder`], a [`wgpu::TextureView`] from the swapchain which you can use to render to the screen, and the default [`ScalingRenderer`].
```
parasyte, line 189: Please split long lines so they do not cross 100 characters. I would also like to keep the initial doc line that describes succinctly what this method does.

src/lib.rs
```diff
@@ -192,21 +181,42 @@ impl Pixels {
...
+    ///     pixel[0] = 0x00; // R
+    ///     pixel[1] = 0x00; // G
+    ///     pixel[2] = 0x00; // B
+    ///     pixel[3] = 0xff; // A
+    /// }
+    ///
+    /// // Draw it to the `SurfaceTexture`
+    /// pixels.render(|encoder, render_target, scaling_renderer| {
+    ///    scaling_renderer.render(encoder, render_target);
+    /// });
```
parasyte, line 215: This doc test is missing a closing

    ```

src/lib.rs
```diff
@@ -192,21 +181,42 @@ impl Pixels {
...
+    pub fn render<F: FnOnce(&mut wgpu::CommandEncoder, &wgpu::TextureView, &ScalingRenderer)>(
```
parasyte, line 216: I'm thinking this `FnOnce` type can be named and defined elsewhere to increase the readability of this function signature.

src/lib.rs
```diff
@@ -224,7 +234,6 @@ impl Pixels {
         });
         mapped.data.copy_from_slice(&self.pixels);
         let buffer = mapped.finish();
-
```
parasyte, line 227: I really miss this empty line. 😢

src/lib.rs
```diff
@@ -371,6 +355,21 @@ impl Pixels {
             pos.1.max(0).min(self.texture_extent.height as isize - 1) as usize,
         )
     }
+
+    /// Provides access the the [`wgpu::Device`] pixels uses.
```
parasyte, line 359: In this three doc comments, we should either put `pixels` in backticks to refer to the crate, or reword the documentation so it doesn't need to refer to itself by name.

src/renderers.rs
```diff
@@ -1,44 +1,37 @@
-use std::fmt;
-use std::rc::Rc;
-use ultraviolet::Mat4;
-use wgpu::{self, Extent3d, TextureView};
-
 use crate::include_spv;
-use crate::render_pass::{BoxedRenderPass, Device, Queue, RenderPass};
+use ultraviolet::Mat4;
+use wgpu::*;
```
parasyte, line 3: Please don't import everything from `wgpu`, there are hundreds of exports in that crate. We only use two dozen or so. Name them all explicitly.

Or you can import only the few that are used multiple times, and undo the removal of the full path from most lines in this file for the one-offs.

examples/conway/src/main.rs
JMS55: As in make render() take Option<FnOnce>, with None corresponding to this? Sure.

src/lib.rs
JMS55: I usually put modules above imports, so main/lib.rs can reference those instead of going through `crate::`. I can change it back.

src/lib.rs
JMS55: I wanted to provide an accessor for it, and I figured it makes sense to not just name it the generic `texture`. I can either remove the prefix, or add it to everything. Your call.

src/lib.rs
JMS55: Sure.

src/lib.rs
JMS55: Good catch.

src/lib.rs
JMS55: Sure.

src/lib.rs
JMS55: Ok xD.

src/lib.rs
JMS55: Sure.

src/renderers.rs
JMS55: Sure.

src/lib.rs
parasyte: I guess I prefer to keep the original name; this field is already a member of a struct named `Pixels`. The redundancy doesn't really add anything, IMHO.

PR #112, opened 2020-08-07
Render API v2

parasyte:
I think this might still need some cleanup in documentation and so on. But this is ready to have more eyes on it.


src/lib.rs
```diff
@@ -551,10 +580,12 @@ impl<'req> PixelsBuilder<'req> {
     /// The default value is [`wgpu::TextureFormat::Rgba8UnormSrgb`], which is 4 unsigned bytes in
     /// `RGBA` order using the SRGB color space. This is typically what you want when you are
     /// working with color values from popular image editing tools or web apps.
-    pub const fn texture_format(
+    ///
+    /// Compressed texture formats are not supported.
```
JMS55, line 584: This is no longer true.

Issue #107, opened 2020-07-22
Render-api-v2 further enhancements


JMS55:
`Pixel` should store and give out `Rc<Device>` and probably `Rc<Queue>` as well.

PixelBuilder should have a `scaling_renderer_with_input_texture(texture: &TextureView)` method, or don't store it in Pixel, and have the user construct it themselves.

JMS55:
I'm conflicted on &Device vs Rc<Device> vs Arc<Device> honestly. I wish there was a good way to let the user choose what they need.

parasyte:
I was thinking about this on the plane the other day. There is no requirement for `pixels` to own the `wgpu` context. It just makes using the library very easy by hiding much of the boilerplate.

The `wgpu` context can borrow structs owned by anything else. It's just a matter of keeping sane lifetimes. This would be exposed through a new method, like `pixels::new_with_context`. I'll try sketching the API sometime.

JMS55:
That also works. I just ran into an issue where I need Arc<Device>, so this would be appreciated.

Also note that https://github.com/parasyte/pixels/blob/wip/render-api-v2/src/renderers.rs#L16 should be &Device

PR #113, opened 2020-08-17
Initial port to wgpu master (0.6)

JMS55:
Surface creation is broken (see examples). I'm not sure how to fix this. Maybe pixels needs to be responsible for creating the surface? Or return the `wgpu::Instance` or something.

I wasn't sure what to do with compressed textures, so I made them cause a panic.

I changed dependencies in Cargo.toml to be in the form `X.Y` instead of `X.Y.Z`. Please correct me if this is wrong.

I have not tested this yet, as the Surface issue needs to be resolved first.

---

Not longer relevant after these changes, but I noticed `ScalingRenderer::resize()` took `&mut wgpu::Device`.


parasyte:
`wgpu::Surface` can be created with [`wgpu::Instance::create_surface()`](https://github.com/gfx-rs/wgpu-rs/blob/02c1fa563f709bfa7b4781b938781bed57069ad1/src/lib.rs#L1297-L1309). Which is only present in `PixelsBuilder::build()` and used to request an adapter. In fact, you can see that the `Surface` is also referenced here to request the adapter: https://github.com/parasyte/pixels/blob/807acc328a5dec604e5dfae05be5e9187992943b/src/lib.rs#L580-L593

`SurfaceTexture` can made generic over any type that implements `HasRawWindowHandle`, and hold a reference to the window passed. Alternatively we can remove or reduce `SurfaceTexture` to just hold a window size, and pass the `HasRawWindowHandle` to the `Pixels::new()` and `PixelsBuilder::build()` methods. Either of these should be enough to defer creating `wgpu::Surface` until the `wgpu::Instance` is available. I like that `SurfaceTexture` keeps everything together that is related to the window, but it is not necessary.

Yet another option is inverting the API so that `Pixels` becomes a `wgpu` middleware instead of the owner. Something like the idea thrown around in https://github.com/parasyte/pixels/issues/107#issuecomment-662656023

parasyte:
I got the `SurfaceTexture` thing resolved on a local branch. I just used the first idea where `SurfaceTexture` is now generic over `HasRawWindowHandle`. I also split it into two structs so `Pixels` can own the size separate from the window reference (which only needs to live as long as the builder).

After I first got everything working, a new panic appeared that I hadn't seen before:

```
thread 'main' panicked at 'the given sampler is/is not a comparison sampler, while the layout type indicates otherwise', /Users/parasyte/.cargo/git/checkouts/wgpu-rs-40ea39809c03c5d8/02c1fa5/src/backend/direct.rs:1323:35
stack backtrace:
   0: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
   1: core::fmt::write
   2: std::io::Write::write_fmt
   3: std::panicking::default_hook::{{closure}}
   4: std::panicking::default_hook
   5: std::panicking::rust_panic_with_hook
   6: rust_begin_unwind
   7: std::panicking::begin_panic_fmt
   8: <core::result::Result<T,E> as wgpu::backend::direct::PrettyResult<T>>::unwrap_pretty::{{closure}}
   9: core::result::Result<T,E>::unwrap_or_else
  10: <core::result::Result<T,E> as wgpu::backend::direct::PrettyResult<T>>::unwrap_pretty
  11: <wgpu::backend::direct::Context as wgpu::Context>::device_create_bind_group
  12: wgpu::Device::create_bind_group
  13: pixels::renderers::ScalingRenderer::new
  14: pixels::PixelsBuilder<W>::build
  15: pixels::Pixels<W>::new
  16: minimal_winit::main
  17: std::rt::lang_start::{{closure}}
  18: std::rt::lang_start_internal
  19: std::rt::lang_start
  20: main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

(Line numbers are missing... This sometimes happens on macOS and I still don't know why.)

The issue is that your patch includes a sampler compare function:

https://github.com/parasyte/pixels/pull/113/files#diff-eb9c4c22014afee4363dad3772426967R34

But the compare function is not actually used, so `comparison` is set to false when creating the binding for the sampler. Changing the linked line to `None` fixed it.

I'm still not satisfied with the update because `wgpu::Instance::create_surface()` is unsafe, and it requires removing `#![forbid(unsafe_code)]`. I feel like moving this to a new private crate that wraps this call in a safe function is probably a bad idea, but also probably better than allowing `unsafe` code in `Pixels`.

parasyte:
For compressed textures, I found the tables in [this Microsoft doc](https://docs.microsoft.com/en-us/windows/win32/direct3d11/texture-block-compression-in-direct3d-11) most helpful. The compression algorithm itself is described in a page linked at the top, but there's also the [wikipedia page](https://en.wikipedia.org/wiki/S3_Texture_Compression) and [this blog post](http://www.reedbeta.com/blog/understanding-bcn-texture-compression-formats/) for additional details.

The TL;DR is all of these compressed formats turn a 4x4 chunk of pixels into either 8 or 16 bytes. So the size is easily computed as `width * height / 16 * i` where `i` is either 8 or 16.

PR #123, opened 2020-10-27
Expose new API method render_texture_format().

schnippl0r:
This method is needed to let the user of the API configure the texture format of the target texture/render texture that the surface texture is rendered on. This texture format is hardware/platform dependent.

For example, this method makes it possible to use the pixels crate on Android because Android seems to not use the texture format `wgpu::TextureFormat::Bgra8UnormSrgb` that is currently hard-coded in `pixels`. Trying to use the current version of `pixels` on Android results in an `adb logcat` entry similar to:

```
requested format Bgra8Srgb is not in list of supported formats:
   [Rgba8Unorm, Rgba8Srgb, R5g6b5Unorm, A2b10g10r10Unorm, Rgba16Sfloat]
```

Using the new method `render_texture_format(TextureFormat::Rgba8UnormSrgb)`, the application starts and displays the pixel buffer as expected. The chosen `TextureFormat` might be hardware dependent but was tested successfully on a Samsung Galaxy S10+ and a Google Pixel 3a.

The name "`render_texture_format`" is open for discussion, as I am not versed enough in graphics programming to determine whether this is the right nomenclature.

A working example (though a bit complicated to set up) using `ndk-glue`, `winit` and my fork of pixels (that implements the new `render_texture_format()` method) can be found at http://github.com/schnippl0r/pixelfire-android

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -92,6 +92,7 @@ pub struct Pixels<W: HasRawWindowHandle> {
     context: PixelsContext,
     surface_size: SurfaceSize,
     present_mode: wgpu::PresentMode,
+    render_texture_format: wgpu::TextureFormat,
     _phantom: std::marker::PhantomData<W>,
 
     // Pixel buffer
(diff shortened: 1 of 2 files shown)


parasyte (approved):

parasyte:
Thanks for the PR!

PR #124, opened 2020-11-26
Update invaders example with Wayland support

reedlaw:
These are the changes I had to make to get this to compile and run on Wayland.

After updating the crates I got a couple errors:
```
error[E0308]: mismatched types
  --> examples/invaders/src/main.rs:63:25
   |
63 |         if input.update(event) {
   |                         ^^^^^
   |                         |
   |                         expected reference, found enum `winit::event::Event`
   |                         help: consider borrowing here: `&event`
   |
   = note: expected reference `&winit::event::Event<'_, _>`
                   found enum `winit::event::Event<'_, ()>`

error[E0599]: no method named `size` found for enum `Option<winit::monitor::MonitorHandle>` in the current scope
   --> examples/invaders/src/main.rs:145:45
    |
145 |         let size = window.current_monitor().size();
    |                                             ^^^^ method not found in `Option<winit::monitor::MonitorHandle>`                                    
```

### DIFF
--- a/examples/invaders/Cargo.toml
+++ b/examples/invaders/Cargo.toml
@@ -18,5 +18,5 @@ log = "0.4.8"
 pixels = { path = "../.." }
 randomize = "3.0.1"
 simple-invaders = { path = "simple-invaders" }
-winit = "0.22.0"
-winit_input_helper = "0.6.0"
+winit = "0.23.0"
+winit_input_helper = "0.8.0"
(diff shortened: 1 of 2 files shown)


parasyte:
@reedlaw Thanks for the PR. I pushed a commit that fixes the lint checks in untouched code paths. You can rebase or merge those changes into your branch to solve those CI failures.

I would like to update the `winit` and `winit_input_helper` dependencies across the entire workspace. Not just this one example.

The last thing is that these examples should probably not be trying to scale the window to fit the screen resolution. This has proven troublesome in the past with Wayland. See: #29.

parasyte:
FWIW, #130 is taking care of updating all of the dependencies. And the lint failure was addressed in #126. That should resolve the issues you experienced. Thanks again for the PR!

PR #134, opened 2021-02-03
Upgrade to wgpu 0.7

JMS55:
TODO:
1. `builder.rs get_texture_format_size()` needs to be updated to support the new formats (replace the `todo!()`). I don't know how to do this myself.
2. Convert shaders to wgsl? (Not neccesary, might not even work, but could be interesting)


JMS55:
Looks like the CI uses Rust 1.43, but wgpu needs a newer version. I'm running 1.49 on my computer and that works.

parasyte:
1.43 is the Minimum Supported Rust Version that works with `pixels` 0.3. We can find the next MSRV for the new dependency tree and update it for CI. It can be a pain, but the process is basically installing versions using a bisect search until you find the oldest one that will work.

Thanks for this BTW. I have another change off-tree to follow up with that replaces the SPIR-V shaders with WGSL.

JMS55:
Reminder to self: I need to finish and push my commit to use `wgpu::Adapter::get_swap_chain_preferred_format()`.

parasyte:
@JMS55 Please rebase or merge the latest HEAD to pick up some CI fixes. The new MSRV may also resolve the issues reported.

JMS55:
I'm a bit confused about what happened with imgui-wgpu, but I believe I've gotten the new commits pulled. The imgui example breaks because it doesn't support wgpu 0.7, but the rest work.

parasyte:
> The imgui example breaks because it doesn't support wgpu 0.7

Yeah, that's expected, unfortunately. There's an open PR for `ingui-wgpu` that is needed: https://github.com/Yatekii/imgui-wgpu-rs/pull/44

JMS55:
On second thought, I'm not sure it makes sense to use `wgpu::Adapter::get_swap_chain_preferred_format()`. Pixels expects the user to write to the buffer in the correct order to match the `TextureFormat` used. Is there a reason we can't just use Bgra8UnormSrgb everywhere? What is the preferred format on the differing platforms?

parasyte:
IIUC, it should be ok to use `wgpu::Adapter::get_swap_chain_preferred_format()` because the texture format for the surface is almost always going to be different than the input texture format. The two texture formats can be specified on the builder: https://github.com/parasyte/pixels/blob/f845d59213acc60e0c0bd84cd60f3413fbd2446b/src/builder.rs#L138-L162

That would be a nice feature improvement... `pixels` could auto-detect the surface texture format, but allow callers to set it manually (as done today with `Builder::render_texture_format`

parasyte:
https://github.com/Yatekii/imgui-wgpu-rs/pull/46 was just merged, so we can probably expect a new release for that dep soon. In the meantime, pointing it at the merge commit might make CI green.

JMS55:
Ok, I updated imgui-wgpu. However we're still failing CI, apparently the minimum version is actually 1.48 due to futures-rs. Anyways once that's fixed I'm calling this PR done. Here's the remaining things that I believe should be handled in separate PR's:
1. Update `builder.rs get_texture_format_size()` to support new formats.
2. Default to `wgpu::Adapter::get_swap_chain_preferred_format()`, and expose the chosen option. I started on this, but I need to put some more thought into it. I forgot that the format you provide your data as to pixels and the format `pixels::ScalingRenderer` uses aren't necessarily the same. Also needs some thought into how we want this to work in the cases where you're using `pixels::ScalingRenderer` via `pixels::render_with()`. It also might require a bit of refactoring. I definitely need to think on this longer.
3. Port shaders to WGSL (I wouldn't mind handling this either, but @parasyte you said you were working on this already).

parasyte:
> However we're still failing CI, apparently the minimum version is actually 1.48 due to futures-rs.

I see. This is actually the future readiness feature in `std`, which was stabilized in 1.48, and is required by wgpu 0.7. We should definitely be making that change to fix CI before merging this PR.

WGSL shaders are a WIP. I'm happy to open a PR when I can rewrite it on top of this work.

parasyte (approved):
Thanks @JMS55 for updating `wgpu`. I have merged recent changes, added support for the new compressed texture formats, documented how texture sizes are computed, and updated the MSRV so CI passes.

I'll create followup tickets for the outstanding items you listed.

Issue #125, opened 2020-12-10
How can I resize the pixel buffer itself


clay53:
I want to be able to change the resolution the program is running at after initialization. I'm aware that `Pixels.resize(width: u8, height: u8)` resizes the surface the buffer is drawn on but that's not what I'm looking for. I've tried creating a new instance of pixels and replacing the old one like this: `pixels = pixels::PixelsBuilder::new(window_size.width, window_size.height, SurfaceTexture::new(window_size.width, window_size.height, &window)).enable_vsync(false).build().unwrap();`. However, this causes an error when attempting to render it again: [log.txt](https://github.com/parasyte/pixels/files/5673970/log.txt)

parasyte:
The error is the same thing described in #121. There is a PR you can try in #122 It is not merge because there is an open question: https://github.com/parasyte/pixels/issues/121#issuecomment-719968913

Other than that, there isn't a method available for resizing the pixel buffer, just because I didn't need this for emulators of old hardware (.e.g with a fixed screen resolution). But I'm happy to accept contributions for a feature.

clay53:
> The error is the same thing described in #121. There is a PR you can try in #122 It is not merge because there is an open question: [#121 (comment)](https://github.com/parasyte/pixels/issues/121#issuecomment-719968913)
> 
> Other than that, there isn't a method available for resizing the pixel buffer, just because I didn't need this for emulators of old hardware (.e.g with a fixed screen resolution). But I'm happy to accept contributions for a feature.

#122 doesn't seem to be working for what I want. I get a segmentation fault when I resize the window replacing the old instance of pixels to resize the pixel buffer. It does work when I do `pixels.resize()` but that was already working fine on my Linux machine which, I think, is what people are talking about isn't working in #121.

parasyte:
To be clear, I was not suggesting that #122 would help. Just observing that the error is the same. #121 is about resizing the window and `wgpu` returning an error due to some race conditions.

Also looking over the attached log again, and it's clear that it doesn't really have enough context to troubleshoot. If you would like to provide a more complete example source, I think it can be useful to see what has already been tried before.

LoganDark:
This is a huge deal breaker for me. `pixels` seemed like a good candidate to implement a fantasy computer, but if it does not support resizing the screen, I will have to look elsewhere (minifb?).

parasyte:
I added the `help wanted` label to advertise that this feature is up for anyone who would like to implement it.

PR #122, opened 2020-10-27
Re-create swapchain in case it becomes out of date (fixes #121)

svenstaro:

Issue #121, opened 2020-10-09
Applications exit/crash on window resize on Linux systems


schnippl0r:
Hi @parasyte,

I have written a small application using pixels and found out that the application crashes on a window resize. I can reproduce the same behaviour with running the example application `minimal-winit`.

Resizing the window provides the following error message:
`[2020-10-09T18:56:49Z ERROR minimal_winit] pixels.render() failed: The GPU failed to acquire a swapchain frame.`

Removing the error handling on the `pixels.render()` call (line 49) and just `unwrap()`-ing it delivers the following error message:
`thread 'main' panicked at 'called 'Result::unwrap()' on an 'Err' value: Swapchain(Outdated)', examples/minimal-winit/src/main.rs:49:29`

I tested the window resize on different Linux systems and graphics drivers: ArchLinux+amdgpu, ArchLinux+nvidia, Debian 10+intel. All systems were running X11, not Wayland.
Performing a window resize with the example running under Windows 10 is working as expected.

I have also tested other examples from frameworks that base on winit (piston, amethyst) and window resizing works as expected.

Kind regards

Full Backtrace:
```
thread 'main' panicked at 'called 'Result::unwrap()' on an 'Err' value: Swapchain(Outdated)', examples/minimal-winit/src/main.rs:49:29
stack backtrace:
   0:     0x562710bc8340 - std::backtrace_rs::backtrace::libunwind::trace::h577ea05e9ca4629a
                               at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/../../backtrace/src/backtrace/libunwind.rs:96
   1:     0x562710bc8340 - std::backtrace_rs::backtrace::trace_unsynchronized::h50b9b72b84c7dd56
                               at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/../../backtrace/src/backtrace/mod.rs:66
   2:     0x562710bc8340 - std::sys_common::backtrace::_print_fmt::h6541cf9823837fac
                               at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/sys_common/backtrace.rs:79
   3:     0x562710bc8340 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hf64fbff071026df5
                               at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/sys_common/backtrace.rs:58
[...]

PR #136, opened 2021-02-05
Resize buffer (issue #125)

JoshLambda:
Function to resize the pixel buffer:

pub fn resize_buffer(&mut self, width: u32, height: u32)

The backing texture and swap chain are re-created and the buffer is resized.
The texture format also has to be stored at creation of `Pixels` so that the texture can be re-created with the same format.

I also updated example `imgui-winit` to account for the resized width when `world::draw()`-ing.

This ~~addresses~~ may help with #125.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,4 +1,5 @@
 use crate::renderers::{ScalingMatrix, ScalingRenderer};
+use crate::SurfaceSize;
 use crate::{Error, Pixels, PixelsContext, SurfaceTexture};
 use raw_window_handle::HasRawWindowHandle;
 use std::env;
@@ -188,33 +189,6 @@ impl<'req, 'win, W: HasRawWindowHandle> PixelsBuilder<'req, 'win, W> {
             pollster::block_on(adapter.request_device(&self.device_descriptor, None))
                 .map_err(Error::DeviceNotFound)?;
 
-        // The rest of this is technically a fixed-function pipeline... For now!
-
-        // Create a texture
-        let width = self.width;
-        let height = self.height;
-        let texture_extent = wgpu::Extent3d {
-            width,
-            height,
-            depth: 1,
-        };
-        let texture = device.create_texture(&wgpu::TextureDescriptor {
-            label: Some("pixels_source_texture"),
-            size: texture_extent,
-            mip_level_count: 1,
-            sample_count: 1,
-            dimension: wgpu::TextureDimension::D2,
-            format: self.texture_format,
-            usage: wgpu::TextureUsage::SAMPLED | wgpu::TextureUsage::COPY_DST,
-        });
-        let texture_view = texture.create_view(&wgpu::TextureViewDescriptor::default());
-        let texture_format_size = get_texture_format_size(self.texture_format);
-
-        // Create the pixel buffer
-        let capacity = ((width * height) as f32 * texture_format_size) as usize;
-        let mut pixels = Vec::with_capacity(capacity);
-        pixels.resize_with(capacity, Default::default);
-
         let present_mode = self.present_mode;
 
         // Create swap chain
@@ -223,25 +197,28 @@ impl<'req, 'win, W: HasRawWindowHandle> PixelsBuilder<'req, 'win, W> {
             &mut device,
             &surface,
             self.render_texture_format,
-            surface_size.width,
-            surface_size.height,
+            &surface_size,
             present_mode,
         );
 
-        let scaling_matrix_inverse = ScalingMatrix::new(
-            (width as f32, height as f32),
-            (surface_size.width as f32, surface_size.height as f32),
-        )
-        .transform
-        .inversed();
+        // Create the backing texture
+        let (scaling_matrix_inverse, texture_extent, texture, scaling_renderer, pixels_buffer_size) =
+            create_backing_texture(
+                &device,
+                // Backing texture values
+                self.width,
+                self.height,
+                self.texture_format,
+                // Render texture values
+                &surface_size,
+                self.render_texture_format,
+            );
 
-        let scaling_renderer = ScalingRenderer::new(
-            &device,
-            &texture_view,
-            &texture_extent,
-            self.render_texture_format,
-        );
+        // Create the pixel buffer
+        let mut pixels = Vec::with_capacity(pixels_buffer_size);
+        pixels.resize_with(pixels_buffer_size, Default::default);
 
+        // Instantiate the Pixels struct
         let context = PixelsContext {
             device,
             queue,
@@ -250,9 +227,10 @@ impl<'req, 'win, W: HasRawWindowHandle> PixelsBuilder<'req, 'win, W> {
             texture,
             texture_extent,
             texture_format: self.texture_format,
-            texture_format_size,
+            texture_format_size: get_texture_format_size(self.texture_format),
             scaling_renderer,
         };
+
         Ok(Pixels {
             context,
             surface_size,
@@ -268,8 +246,9 @@ pub(crate) fn create_swap_chain(
     device: &mut wgpu::Device,
     surface: &wgpu::Surface,
     format: wgpu::TextureFormat,
-    width: u32,
-    height: u32,
+    surface_size: &SurfaceSize,
+    // width: u32,
+    // height: u32,
     present_mode: wgpu::PresentMode,
 ) -> wgpu::SwapChain {
     device.create_swap_chain(
@@ -277,14 +256,72 @@ pub(crate) fn create_swap_chain(
         &wgpu::SwapChainDescriptor {
             usage: wgpu::TextureUsage::OUTPUT_ATTACHMENT,
             format,
-            width,
-            height,
+            width: surface_size.width,
+            height: surface_size.height,
             present_mode,
         },
     )
 }
 
-fn get_texture_format_size(texture_format: wgpu::TextureFormat) -> f32 {
+pub(crate) fn create_backing_texture(
+    device: &wgpu::Device,
+    width: u32,
+    height: u32,
+    backing_texture_format: wgpu::TextureFormat,
+    surface_size: &SurfaceSize,
+    render_texture_format: wgpu::TextureFormat,
+) -> (
+    ultraviolet::Mat4,
+    wgpu::Extent3d,
+    wgpu::Texture,
+    ScalingRenderer,
+    usize,
+) {
+    let scaling_matrix_inverse = ScalingMatrix::new(
+        (width as f32, height as f32),
+        (surface_size.width as f32, surface_size.height as f32),
+    )
+    .transform
+    .inversed();
+
+    let texture_extent = wgpu::Extent3d {
+        width,
+        height,
+        depth: 1,
+    };
+
+    let texture = device.create_texture(&wgpu::TextureDescriptor {
+        label: Some("pixels_source_texture"),
+        size: texture_extent,
+        mip_level_count: 1,
+        sample_count: 1,
+        dimension: wgpu::TextureDimension::D2,
+        format: backing_texture_format,
+        usage: wgpu::TextureUsage::SAMPLED | wgpu::TextureUsage::COPY_DST,
+    });
+    let texture_view = texture.create_view(&wgpu::TextureViewDescriptor::default());
+
+    let scaling_renderer = ScalingRenderer::new(
+        device,
+        &texture_view,
+        &texture_extent,
+        render_texture_format,
+    );
+
+    let texture_format_size = get_texture_format_size(backing_texture_format);
+    let pixels_buffer_size = ((width * height) as f32 * texture_format_size) as usize;
+
+    (
+        scaling_matrix_inverse,
+        texture_extent,
+        texture,
+        scaling_renderer,
+        pixels_buffer_size,
+    )
+}
+
+#[inline]
+const fn get_texture_format_size(texture_format: wgpu::TextureFormat) -> f32 {
     match texture_format {
         // 8-bit formats
         wgpu::TextureFormat::R8Unorm
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -251,19 +251,35 @@ impl<'req, 'win, W: HasRawWindowHandle> PixelsBuilder<'req, 'win, W> {
             swap_chain,
             texture,
             texture_extent,
+            texture_format: self.texture_format,
             texture_format_size,
             scaling_renderer,
         };
-
-        Ok(Pixels {
+        let mut pixels = Pixels {
             context,
             surface_size,
             present_mode,
             pixels,
             scaling_matrix_inverse,
             render_texture_format: self.render_texture_format,
-        })
+        };
+        create_swap_chain(&mut pixels);
+
+        Ok(pixels)
     }
+}
+
+pub(crate) fn create_swap_chain(pixels: &mut Pixels) {
+    pixels.context.swap_chain = pixels.context.device.create_swap_chain(
+        &pixels.context.surface,
+        &wgpu::SwapChainDescriptor {
+            usage: wgpu::TextureUsage::OUTPUT_ATTACHMENT,
+            format: pixels.render_texture_format,
+            width: pixels.surface_size.width,
+            height: pixels.surface_size.height,
+            present_mode: pixels.present_mode,
+        },
+    );
 }
 
 fn get_texture_format_size(texture_format: wgpu::TextureFormat) -> f32 {
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -184,7 +184,7 @@ impl<'req, 'win, W: HasRawWindowHandle> PixelsBuilder<'req, 'win, W> {
         ));
         let adapter = pollster::block_on(adapter).ok_or(Error::AdapterNotFound)?;
 
-        let (device, queue) =
+        let (mut device, queue) =
             pollster::block_on(adapter.request_device(&self.device_descriptor, None))
                 .map_err(Error::DeviceNotFound)?;
 
@@ -219,15 +219,13 @@ impl<'req, 'win, W: HasRawWindowHandle> PixelsBuilder<'req, 'win, W> {
 
         // Create swap chain
         let surface_size = self.surface_texture.size;
-        let swap_chain = device.create_swap_chain(
+        let swap_chain = create_swap_chain(
+            &mut device,
             &surface,
-            &wgpu::SwapChainDescriptor {
-                usage: wgpu::TextureUsage::OUTPUT_ATTACHMENT,
-                format: self.render_texture_format,
-                width: surface_size.width,
-                height: surface_size.height,
-                present_mode,
-            },
+            self.render_texture_format,
+            surface_size.width,
+            surface_size.height,
+            present_mode,
         );
 
         let scaling_matrix_inverse = ScalingMatrix::new(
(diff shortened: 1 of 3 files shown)


LoganDark:
I don't think this is the right solution. `pixels` should not automatically choose a new buffer size for you; you should be the one to specify.

JoshLambda:
> I don't think this is the right solution. `pixels` should not automatically choose a new buffer size for you; you should be the one to specify.

The buffer is copied to the texture no matter what.
It can only be te same size or smaller (rendered with black bands).

If you want a different buffer size, you have to create your own as an off-screen rendering target.
Then it's up to you to copy/scale the part you want to the actual rendering buffer.
This crate could handle that for you, but I'm not sure that's the goal of a simple framebuffer lib.

LoganDark:
> > I don't think this is the right solution. `pixels` should not automatically choose a new buffer size for you; you should be the one to specify.
> 
> The buffer is copied to the texture no matter what.
> It can only be te same size or smaller (rendered with black bands).
> 
> If you want a different buffer size, you have to create your own as an off-screen rendering target.
> Then it's up to you to copy/scale the part you want to the actual rendering buffer.
> This crate could handle that for you, but I'm not sure that's the goal of a simple framebuffer lib.

A simple framebuffer lib should give you the ability to specify the pixel dimensions of the framebuffer. According to your description, the pixel dimensions of the framebuffer are automatically decided when the window resizes. However, `pixels` is meant to have an **explicit** resolution.

examples/imgui-winit/Cargo.toml
```diff
@@ -16,5 +16,5 @@ imgui-wgpu = "0.12"
 imgui-winit-support = "0.6"
 log = "0.4"
 pixels = { path = "../.." }
-winit = "0.24"
-winit_input_helper = "0.9"
+winit = "0.23"
+winit_input_helper = "0.8"
```
LoganDark, line 20: Hm, did the later versions of these introduce some sort of regression or unwanted behavior?

JoshLambda:
> A "simple framebuffer lib" as you describe should give you the ability to specify the pixel dimensions of the framebuffer. According to your description, the pixel dimensions of the framebuffer are automatically decided when the window resizes. However, `pixels` is meant to have an **explicit** resolution.
> 
> Please note that the pixel dimensions of the framebuffer are usually _different_ than the physical dimensions of the surface or swap chain, which must be kept in sync with the window, and do not need to be explicitly specified by the programmer.

I agree, but `PixelsBuilder::pixel_aspect_ratio` is currently hardcoded to 1.0 simply because handling a different aspect ratio is not implemented yet. The point of this PR was to be able to resize the content of the window with the window itself (assuming a pixel ratio of 1.0), which is an improvement to the current padding with black bands.

When resizing the window, the buffer size is computed this way:  `capacity = width * height * texture_format_size`
You typically get `width` and `height` dimensions in physical pixels using `window.inner_size()`

So if you want to, you could apply some ratio yourself to `width` and `height` when calling `pixels.resize()`, but I'm not sure it will do what you want it to do.

Maybe if you could give a specific use case I would get a better idea.

LoganDark:
> > A "simple framebuffer lib" as you describe should give you the ability to specify the pixel dimensions of the framebuffer. According to your description, the pixel dimensions of the framebuffer are automatically decided when the window resizes. However, `pixels` is meant to have an **explicit** resolution.
> > Please note that the pixel dimensions of the framebuffer are usually _different_ than the physical dimensions of the surface or swap chain, which must be kept in sync with the window, and do not need to be explicitly specified by the programmer.
> 
> I agree, but `PixelsBuilder::pixel_aspect_ratio` is currently hardcoded to 1.0 simply because handling a different aspect ratio is not implemented yet. The point of this PR was to be able to resize the content of the window with the window itself (assuming a pixel ratio of 1.0), which is an improvement to the current padding with black bands.
> 
> When resizing the window, the buffer size is computed this way: `capacity = width * height * texture_format_size`
> You typically get `width` and `height` dimensions in physical pixels using `window.inner_size()`
> 
> So if you want to, you could apply some ratio yourself to `width` and `height` when calling `pixels.resize()`, but I'm not sure it will do what you want it to do.
> 
> Maybe if you could give a specific use case I would get a better idea.

I'm not talking about aspect ratios either. Pixels in `pixels` are always square, but the framebuffer does not have to have the same size as the window. Yes, the black borders you mention come up when the aspect ratio of the buffer does not match up with the window; this does not need to change. What I think should change is the fact that currently `pixels` decides the new resolution of the framebuffer for you, rather than letting you decide yourself.

What if you want to resize the buffer but the window size has not changed? For example, an emulator with multiple different graphics modes (which is quite common). Right now you have to make `pixels` think the window size has changed, which is not a great solution.

examples/imgui-winit/Cargo.toml
JoshLambda: Yes, the example wouldn't build at all, because winit_input_helper 0.9 requires winit 0.24, while other crates require winit 0.23. So I had to downgrade both of them to avoid a version conflict.

JoshLambda:
> I'm not talking about aspect ratios either. Pixels in `pixels` are always square, but the framebuffer does not have to have the same size as the window. Yes, the black borders you mention come up when the aspect ratio of the buffer does not match up with the window; this does not need to change. What I think should change is the fact that, in your PR, `pixels` will decide the new resolution of the framebuffer for you, rather than letting you decide yourself.
> 
> What if you want to resize the buffer but the window size has not changed? For example, an emulator with multiple different graphics modes (which is quite common). Right now you have to make `pixels` think the window size has changed, which is not a great solution.

Ok, now I understand better. `Pixels` offers only a physical frame buffer. Which means the buffer is the exact representation of the physical pixels from the window. 
As far as I understand - I'm not the maintainer - this crate is meant to be minimalistic, so it's is not a convenience to represent your pixels abstractly and then scale them for you back to the physical space. In this context this is your own job to do so.

I know that it would be useful, as there are lots of operations we can do with buffers. 
But there you have 2 choices, either as I said earlier, creating a different buffer yourself as an intermediate render target and then scale and copy the content to the physical buffer at each update. Or find a more abstract and high level library which would suit your needs.

LoganDark:
> As far as I understand - I'm not the maintainer - this crate is meant to be minimalistic, so it's is not a convenience to represent your pixels abstractly and then scale them for you back to the physical space. In this context this is your own job to do so.

Um... but it is though? That's almost the exact purpose of `pixels`.

Quick ping to @parasyte, because I'm not sure if I can clear this up. As-is, I don't think this change fits with the spirit of `pixels`.

parasyte:
Hi, Thanks for the PR @JoshLambda. I'll have to look over it more closely, probably this weekend.

If we're talking about the _original intention_ of the crate, then that's pretty easy: The pixel buffer was always meant to be fixed sized. The target audience was like NES emulator and Chip-8 authors. These machines only have one screen resolution, so a fixed size screen was perfect.

FWIW, the scaling ratio is intended to support non-square pixel aspect ratios like NES's 8:7 PAR. It's currently unimplemented because I'm kind of lazy. It can be ignored for now, but the short explanation is that it can be used to horizontally (or vertically) stretch the backing pixel buffer while maintaining sharp pixel edges. For this reason, it will never scale both horizontally and vertically at the same time. There are some more technical details and example screenshots on my blog: https://blog.kodewerx.org/2017/09/tiny-mcu-3d-renderer-part-6-camera.html This is unrelated to scaling in general...

Which leads us to the _original essence_ of `pixels`. Which is to provide a crate that abstracts all of the GPU work away from authors who would rather focus on poking some raw pixel values into an array and getting them to appear on screen. Scaling and custom shaders are a nice bonus because the NES screen resolution is way too small to enjoy at 1:1. Custom shaders give users the option to use NTSC filters, high-quality scaling filters, and so on.

I, for one, am happy to extend the capabilities of the crate to include features that can resize the backing pixel buffer at runtime. E.g. that would allow emulators for old computers to change their video mode/screen resolution. And from that perspective, this PR does not appear to align with the feature request in #125.

----

The `resize` method is for changing the presentation texture size to match the surface (called the surface texture in our code and docs). This method could probably be renamed to avoid any confusion with resizing the pixel buffer (and the backing texture). Of course, `resize` takes care of ensuring the presented image always has sharp pixels, even when the surface size is not evenly divisible by the pixel buffer size. (It does this by padding the image with a black border.) The other thing handled by `resize` is (or eventually will be) pixel aspect ratios. Both operations are simple scaling passes, but they need to know the surface size to work correctly.

The new method that is needed to change the backing pixel buffer (e.g. change the emulator video mode) will have to update the `pixels.context.texture` and recreate the scaling renderer for it.

Both textures should be independently resizable, and as long as the presentation texture size matches the surface size, everything should Just Work :tm:. After changing the video mode, a user might drag the window handle to change the screen size to something more comfortable, or even change the window inner size to some multiple of the new video mode. In all cases, the independent textures will ensure the following properties hold:

1. The image will always be presented with sharp pixels at integer scales, possibly with a black border.
2. The pixel aspect ratio will remain fixed, regardless of the presentation texture/window size.
3. The resolution of the pixel buffer (and associated backing texture) can be resized at runtime to fit application needs. (This is the new part that is needed.)

Hope that helps to clear things up in terms of "the exact purpose of `pixels`" and what I imagined a fix for #125 would look like.

LoganDark:
FWIW, the crate I'm working on supports custom aspect ratios and scaling, custom buffer alignments (other than centered), and high-quality scaling (antialiased pixel edges) already. I plan on supporting custom shaders (and post-processing pipelines in general), multiple layers with transparency, and HDR (floating-point) buffers. Fun!

examples/imgui-winit/Cargo.toml
parasyte: @JoshLambda I just merged a patch that addresses the dependency issues in the `imgui-winit` example. Please rebase or merge on the latest HEAD to pick up the patch. It will conflict with this change, so you will want to revert.

JoshLambda: @parasyte Done.

JoshLambda:
Ok, so I discovered it only works when PR https://github.com/parasyte/pixels/pull/122 about handling `wgpu::SwapChainError::Outdated` is applied, so I had to include it in this PR.

parasyte:
#122 probably needs to be merged at some point, but right now it copy-pastes code instead of keeping things DRY.

There are a few issues with this PR that need to be resolved though. The first is that it breaks CI because the examples haven't been updated with the breaking API change on the `resize` method.

The second issue is that the `resize` method should not be changed. The state of the PR right now doesn't allow the underlying byte vector and backing texture to be resized independently of the surface and presentation texture (because it overloads the `resize` method). As described earlier, this should be split into two separate methods so that these textures can be resized in isolation of one another. The existing `resize` method should probably be renamed to something else, I don't know, `resize_surface` maybe.

I don't know why recreating the swap chain is necessary for resizing the byte vector and backing texture. You're probably running into something like #121 because you're overloading the window resize. Just a guess.

JoshLambda:
> #122 probably needs to be merged at some point, but right now it copy-pastes code instead of keeping things DRY.

I moved the swap chain creation in it's own function, so it's now DRY. 
It's being used at `Pixels` creation, when resizing the backing texture and when handling outdated swap chain error.
 
> There are a few issues with this PR that need to be resolved though. The first is that it breaks CI because the examples haven't been updated with the breaking API change on the `resize` method.

Yes, I didn't look at other examples yet.
 
> The second issue is that the `resize` method should not be changed. The state of the PR right now doesn't allow the underlying byte vector and backing texture to be resized independently of the surface and presentation texture (because it overloads the `resize` method). As described earlier, this should be split into two separate methods so that these textures can be resized in isolation of one another. The existing `resize` method should probably be renamed to something else, I don't know, `resize_surface` maybe.

Now we have:
```
/// Resize the surface only (behavior unchanged, making this PR backward compatible)

pub fn resize(&mut self, width: u32, height: u32)

/// Resize the buffer and backing texture

pub fn resize_buffer(&mut self, width: u32, height: u32)
```
We can rename them at your convenience.
 
> I don't know why recreating the swap chain is necessary for resizing the byte vector and backing texture. You're probably running into something like #121 because you're overloading the window resize. Just a guess.

Yes, some attempt at rendering seems to happen between the time when the window is resized and the swap chain is re-created. But I don't know how.

src/lib.rs
```diff
@@ -182,6 +183,55 @@ impl Pixels {
...
+        };
+
+        // Update ScalingMatrix for mouse transformation
+        self.scaling_matrix_inverse = renderers::ScalingMatrix::new(
+            (
+                self.context.texture_extent.width as f32,
+                self.context.texture_extent.height as f32,
+            ),
+            (width as f32, height as f32),
+        )
```
parasyte, line 201: Hm, this looks suspicious. It's passing the same size for both the backing texture (first tuple) and surface texture (second tuple). I believe it should be this:

```suggestion
        self.scaling_matrix_inverse = renderers::ScalingMatrix::new(
            (width as f32, height as f32),
            (
                self.surface_size.width as f32,
                self.surface_size.height as f32,
            ),
        )
```

src/lib.rs
```diff
@@ -296,11 +337,24 @@ impl Pixels {
...
+                // See https://github.com/parasyte/pixels/issues/121
+                builder::create_swap_chain(self);
+                self.context
+                    .swap_chain
+                    .get_current_frame()
+                    .map_err(Error::Swapchain)?
+            } else {
+                frame_?
+            }
+        };
```
parasyte, line 357: I bet this would look better with `Result::or_else()`...

```suggestion
        let frame = self
            .context
            .swap_chain
            .get_current_frame()
            .or_else(|err| match err {
                wgpu::SwapChainError::Outdated => {
                    // Recreate the swap chain to mitigate race condition on drawing surface resize.
                    // See https://github.com/parasyte/pixels/issues/121
                    builder::create_swap_chain(self);
                    self.context.swap_chain.get_current_frame()
                }
                err => Err(err),
            })
            .map_err(Error::Swapchain)?;
```

src/builder.rs
```diff
@@ -251,21 +251,37 @@ impl<'req, 'win, W: HasRawWindowHandle> PixelsBuilder<'req, 'win, W> {
...
             render_texture_format: self.render_texture_format,
-        })
+        };
+        create_swap_chain(&mut pixels);
+
+        Ok(pixels)
     }
 }
 
+pub(crate) fn create_swap_chain(pixels: &mut Pixels) {
```
parasyte, line 272: Gut feeling is that I would make this a method of the `Pixels` struct instead of a module-level function on `builder`. Thinking I see why you did it this way, though.

There's a call to `create_swap_chain` further up (line 222) that could be removed _iff_ this was a static method that accepted free arguments instead of an instance method... Kind of like it is right now. And that would be fine.

Other options are changing the type of `Context::swap_chain` to `Option<T>` and then unwrapping it everywhere, or trying to create a half-initialized structure using `unsafe` and [`MaybeUninit`](https://doc.rust-lang.org/nightly/std/mem/union.MaybeUninit.html#initializing-a-struct-field-by-field). I'm not a big fan of `unsafe` though. And having to `unwrap()` the swap chain to use it seems unfortunate. Given that, I think I prefer the static method with free args.

parasyte (commented):
Here's my first review. It looks like this might end up closing #121 and #122, which is a nice bonus!

src/lib.rs
```diff
@@ -182,6 +183,55 @@ impl Pixels {
...
+            (
+                self.context.texture_extent.width as f32,
+                self.context.texture_extent.height as f32,
+            ),
+            (width as f32, height as f32),
+        )
+        .transform
+        .inversed();
+
+        // Recreate the texture
```
parasyte, line 205: The rest of this function looks a lot like code in the builder. This _could_ be DRY'd, too. That might help with maintainability?

src/lib.rs
JoshLambda: We are in function `resize()` which is used to resize the surface when the window is resized. So, `width` and `height` are the dimensions of the surface, not the backing texture.
And indeed, 2 lines earlier we have:

        // Update SurfaceTexture dimensions
        self.surface_size.width = width;
        self.surface_size.height = height;

Unless I'm missing something that means both tuples of the matrix function would be the same in your proposal.

JoshLambda: Good catch, updated.

src/lib.rs
JoshLambda: This looks way better indeed. Updated.

src/builder.rs
JoshLambda: I'm missed the useless call at line 222 (the downsize of coding at 4 AM :) ). I'm looking at it.

JoshLambda: The useless call to `device.create_swap_chain` line 222 has been removed.
Now we have:

- in builder.rs
/// Generic swap chain creation with free arguments
`pub(crate) fn create_swap_chain(device: &mut wgpu::Device, surface: &wgpu::Surface, format: wgpu::TextureFormat,
    width: u32, height: u32, present_mode: wgpu::PresentMode)`

- in lib.rs
/// Method of Pixels, which calls above function with own values
`pub(crate) fn re_create_swap_chain(&mut self) { /* calls create_swap_chain above, with own values */ }`

src/lib.rs
JoshLambda: Yes, that makes sense. I'll look at it when the part about swap chain creation is stabilized.

JMS55:
Since it looks like someone else is tackling this, here's my fork that I never finished. Feel free to use whatever you need from it https://github.com/JMS55/pixels-resize-buffer/commit/b9f8008c87666133784199f8dbbe35804a90fc78.

src/builder.rs
```diff
@@ -266,6 +264,26 @@ impl<'req, 'win, W: HasRawWindowHandle> PixelsBuilder<'req, 'win, W> {
     }
 }
 
+pub(crate) fn create_swap_chain(
+    device: &mut wgpu::Device,
+    surface: &wgpu::Surface,
+    format: wgpu::TextureFormat,
+    width: u32,
+    height: u32,
```
parasyte, line 272: nitpick: This could accept a `SurfaceSize` as a single arg. Both call sites have direct access to one.

src/lib.rs
```diff
@@ -333,6 +383,18 @@ impl Pixels {
         Ok(())
     }
 
+    // Re-create the swap chain with its own values
+    pub(crate) fn re_create_swap_chain(&mut self) {
```
parasyte, line 387: nitpick: minor name change to `recreate_swap_chain` would be preferred.

parasyte (commented):
The latest changes have addressed most of my concerns. I have a few nitpicks left (I'm not opposed to keeping things how they are; your call). And the last bit to take care of is DRYing the texture and scaling renderer creation in `Pixels::resize_buffer` and `Builder::build`.

src/builder.rs
JoshLambda: Good idea, done.

src/lib.rs
JoshLambda: Of course, done.

JoshLambda:
I made the builder and buffer_resize as DRY as I could. How does it look like now?

parasyte (approved):
This is awesome. Thanks for putting this together, and also sorry about the delay.

JMS55:
Just to check, nothing needs to be changed with `ScalingRenderer`? When I was working on this, I was under the impression that you had to change some things (the matrix I think?) whenever the backing texture size changes.

parasyte:
The matrix is for resizing the image on the surface texture. The size of the backing texture should be completely independent of the surface texture or how it gets displayed.

Issue #87, opened 2020-06-09
Black screen on macOS Mojave


FloVanGH:
I've got black screen running the winit examples on macOS Mojave.

![image](https://user-images.githubusercontent.com/6715107/84105794-aa857a00-aa19-11ea-8588-e0c12f32534a.png)

parasyte:
Related to #72 and #76

parasyte:
I have added a [troubleshooting section](https://github.com/parasyte/pixels#troubleshooting) to the README, with instructions for requesting the high performance GPU. This should be enough to workaround this issue for now.

FloVanGH:
After I've updated to macOS Catalina it works.

LoganDark:
Apparently, everyone who's encountered this issue has just promptly updated their OS, preventing the maintainers from ever determining or fixing the problem.

Mojave is only 2 versions behind the latest version. It's also more significant than other versions because it's the first version to add dark mode AND the last version to support 32-bit programs. That means some people like me are never going to upgrade.

That said, I also cannot reproduce this issue. I would love to help, but you know...

parasyte:
The most up-to-date information I have on this issue is that it relates to both the GPU hardware and version of Metal that ships with Mojave. I have seen black screens with integrated Intel GPUs, but not discrete AMD GPUs.

Apart from upgrading the OS, requesting the high performance GPU may be acceptable in most cases. There is still an issue with laptops that only have an Intel GPU and run Mojave, however...

LoganDark:
> There is still an issue with laptops that only have an Intel GPU and run Mojave, however...

...and laptops which turn on automatic graphics switching every time you reboot, like mine.

parasyte:
The last release of Mojave was July 21, 2021; 16 months ago. It has been succeeded by four major macOS releases. I think this can safely be closed and our official position should be "macOS Mojave is not supported".

LoganDark:
What a great way to solve the issue, just wait until you can say the OS is not supported :)

parasyte:
If you have a better solution, I will happily accept it.

Unfortunately there has been a lack of contributions on this particular issue, and I do not personally have a way to run macOS Mojave. And as you know, there are open issues on the `wgpu` repo for Mojave. And at least one that was closed (probably prematurely): https://github.com/gfx-rs/wgpu/issues/104

Regardless, I don't think there is anything we can do here except track an upstream issue (whether that's the closed one or something else?)

Issue #72, opened 2020-04-16
Black screen on macOS 10.14.6 


bthuilot:
When running `pixels` on my MacBook the screen remains black. I've tried running the simple examples on both my MacBook and Linux desktop and everything worked as expected on linux.
 
rust version
`rustc 1.40.0 (73528e339 2019-12-16)`
pixels version
`pixels = "0.0.3"`

parasyte:
It will be good to get to the bottom of this one! I develop primarily on macOS 10.15.4, and haven't seen this behavior.

Can you provide the following info, please:

* macOS version and GPU hardware.
  * E.g. run `system_profiler SPSoftwareDataType SPDisplayDataType` in your terminal and paste the output.
* Info and trace logs:
  * `RUST_LOG=info cargo run --package minimal-winit 2>&1 | head`
  * `RUST_LOG=trace cargo run --package minimal-winit 2>&1 | head -n 1000`

bthuilot:
Output of system_profiler 
```
Software:

    System Software Overview:

      System Version: macOS 10.14.6 (18G3020)
      Kernel Version: Darwin 18.7.0
      Boot Volume: MacHD
      Boot Mode: Normal
      Computer Name: Bryce’s MacBook Pro
      User Name: Bryce Thuilot (bryce)
      Secure Virtual Memory: Enabled
      System Integrity Protection: Disabled
      Time since boot: 5 days 7:23

Graphics/Displays:

    Intel Iris Plus Graphics 640:

      Chipset Model: Intel Iris Plus Graphics 640
      Type: GPU
      Bus: Built-In
      VRAM (Dynamic, Max): 1536 MB
      Vendor: Intel
      Device ID: 0x5926
      Revision ID: 0x0006
      Metal: Supported, feature set macOS GPUFamily2 v1
      Displays:
        Color LCD:
          Display Type: Built-In Retina LCD
          Resolution: 2560 x 1600 Retina
          Framebuffer Depth: 24-Bit Color (ARGB8888)
          Main Display: Yes
          Mirror: Off
          Online: Yes
          Rotation: Supported
          Automatically Adjust Brightness: No
```
Info:
```
 Finished dev [unoptimized + debuginfo] target(s) in 0.38s
     Running `/Users/bryce/build/pixels/target/debug/minimal-winit`
[...]

PR #137, opened 2021-02-06
add fltk example

MoAlyousef:
Hello
Thank you for the nice and easy to use crate. This PR adds an example using the fltk crate. The window passed to Pixels can be the top level window (like in the example) or an embedded window. I took the liberty to change the minimal example to be of a circle bouncing around.
Thanks again


examples/minimal-fltk/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[features]
+optimize = ["log/release_max_level_warn"]
+default = ["optimize"]
+
+[dependencies]
+fltk = "0.14"
+env_logger = "0.7.1"
+log = "0.4.8"
+pixels = { path = "../.." }
```
bjorn3, line 16: nit: missing trailing newline

MoAlyousef: oh. I didn't know a new line was required.

fixed

parasyte:
This looks good to me at first glance. @MoAlyousef please rebase or merge the latest HEAD to pick up an unrelated CI patch.

MoAlyousef:
> This looks good to me at first glance. @MoAlyousef please rebase or merge the latest HEAD to pick up an unrelated CI patch.

Sure

parasyte:
Ok, I tested this on macOS (11.2, Big Sur) and I don't get anything drawn on the fltk window. I also tested the `pong` demo from the fltk repo, and it does work. So there seems to be something missing. Debugging reveals that the closure passed to `win.draw()` does not run at all. Any thoughts?

Also tried this PR on Windows and it works fine there.

MoAlyousef:
Unfortunately I can't get any of the other Pixels demos/examples to run on my MacOS Big Sur (installed on a macbook air 2015). 
Even with the PIXELS_HIGH_PERF env variable set prior to running the build, so I assumed it's just my drivers. Normally however calls to win.redraw() should call the win.draw() callback.

MoAlyousef:
Anyways the win.draw() call is not really necessary. This for example works (on my windows machine):
```rust
fn main() -> Result<(), Error> {
    env_logger::init();
    let app = app::App::default();
    let mut win = Window::default()
        .with_size(WIDTH as i32, HEIGHT as i32)
        .with_label("Hello Pixels");
    win.end();
    win.make_resizable(true);
    win.show();

    let mut pixels = {
        let surface_texture = SurfaceTexture::new(WIDTH, HEIGHT, &win);
        Pixels::new(WIDTH, HEIGHT, surface_texture)?
    };

    let mut world = World::new();

    while app.wait() {
        // Update internal state
        world.update();
        // Draw the current frame
        world.draw(pixels.get_frame());
        pixels.render().unwrap()
        app::redraw();
        // Calls to redraw in the event loop require an explicit sleep
        thread::sleep(Duration::from_millis(SLEEP));
    }

    Ok(())
}
```

parasyte:
> Unfortunately I can't get any of the other Pixels demos/examples to run on my MacOS Big Sur (installed on a macbook air 2015).
> Even with the PIXELS_HIGH_PERF env variable set prior to running the build, so I assumed it's just my drivers. Normally however calls to win.redraw() should call the win.draw() callback.
> 
> Edit: I'll try to investigate things further on a colleague's device.

Thanks. Let me know if there is anything I can help with on the `pixels` side. I do want to make sure the crates works on all modern hardware, but macOS tends to be tricky... #72 #76 #87 and others...

> Anyways the win.draw() call is not really necessary. This for example works (on my windows machine):

Awesome! This gets it a little farther along on macOS. I had to change `app::redraw()` back to `win.redraw()`. The former only causes the window to redraw when the window receives events (like moving the mouse cursor).

MoAlyousef:
I've updated the example and removed the call to win.draw(). FLTK documents that calling Fl_Widget::draw on apple systems are expected to call [Quartz drawing](https://www.fltk.org/doc-1.3/osissues.html#osissues_macos) functions (or those wrapped by FLTK itself). Apparently the code gets optimized out since there are no draw calls. 
`All code inside Fl_Widget::draw() is expected to call Quartz drawing functions.` 
Anyways, I also tried some of the wgpu examples and all give me blank windows on my macbook air. I still haven't had the chance to try things out on my friend's Mac.

parasyte (approved):

parasyte:
Thank you for your contribution! The addition of an example of `fltk` provides more options for developers to use GPU-acceleration in their applications. I appreciate it.

> FLTK documents that calling Fl_Widget::draw on apple systems are expected to call [Quartz drawing](https://www.fltk.org/doc-1.3/osissues.html#osissues_macos) functions (or those wrapped by FLTK itself). Apparently the code gets optimized out since there are no draw calls.
> `All code inside Fl_Widget::draw() is expected to call Quartz drawing functions.`

This is a bit surprising. I'm glad you were able to track that down, though. It seems like an easy mistake to make. I'll remember this for any future issues reported on macOS with `fltk`!

> Anyways, I also tried some of the wgpu examples and all give me blank windows on my macbook air. I still haven't had the chance to try things out on my friend's Mac.

You may want to try the `gfx` examples in addition. It might help narrow down whether the problem exists in `wgpu` or in its lower layers. Did you try the `wgpu` examples for 0.6 or 0.7? We have a release forthcoming with 0.7 support, and it would be nice if that fixed some bugs. 😄

MoAlyousef:
Thank you. 
I've tried wgpu 0.7. The examples worked well building for release on my Windows machine. On MacOS I haven't had luck. I'll try the gfx examples next. 
I kinda wish I hadn't upgraded to Big Sur, every thing now feels kinda buggy and slow. Link times take a little longer than a minute!

parasyte:
> Link times take a little longer than a minute!

I use [`zld`](https://github.com/michaeleisel/zld#if-using-rust) as my default linker. It can still be slow, but is noticeably faster than the native Apple linker. It can be installed with homebrew:

```bash
$ brew tap michaeleisel/zld
$ brew install zld
```

MoAlyousef:
Thanks again. I'll check it out

PR #143, opened 2021-03-01
Fix matrix when creating a scaling renderer

parasyte:
- This issue can be seen when creating a window and pixel buffer that have differing sizes; the image will be stretched to fill the window.
- If the window supports the resize callback to correct the image aspect ratio, resizing the window will correct the matrix immediately, adding the black border as expected. This is a jarring effect when the texture size ratio is not an integer.
- This bug also causes issues with the new `resize_buffer()` API.

This fixes a long-standing issue with the scaling renderer. It was made more obvious with the new `resize_buffer()` API, and reported in https://github.com/parasyte/pixels/pull/136#issuecomment-787532492

r? @JMS55

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -309,6 +309,7 @@ pub(crate) fn create_backing_texture(
         device,
         &texture_view,
         &texture_extent,
+        surface_size,
         render_texture_format,
     );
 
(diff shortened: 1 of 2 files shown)


JMS55:
Looks good to me!

PR #146, opened 2021-03-07
Add `egui` example.

parasyte:
This fails on 1.48 with a feature that was stabilized in 1.49. I don't want to bump the MSRV (yet) just to support an example. I also want to wait for the next `egui_wgpu_backend` release, which supports `egui` 0.10.

... Since I have pulled in the current `egui_wgpu_backend` crate from git, that latter problem is solved. To address the former issue, I'm planning to rewrite the GUI to just create a hello-world window (and probably not much more) instead of using `egui_demo_lib`. That will eliminate need for the `epi` crate, which is where the MSRV bump comes from. Removing these crates also means we can drop the `chrono` dep.


parasyte:
*sigh* `egui` 0.10 uses `bool::then` which was stabilized in Rust 1.50. I guess there is no escaping the MSRV bump.

PR #151, opened 2021-03-10
Implement pixel aspect ratio

parasyte:
This is still missing linear texture filtering to smooth out the stretching artifacts that it introduces. The images shown in the example README show the artifacts that I'm talking about. The pixels are chunkier than they should be. For additional info on how it's _supposed_ to look, [this blog post](https://blog.kodewerx.org/2017/09/tiny-mcu-3d-renderer-part-6-camera.html) goes into plenty of detail. Another resource is https://tanalin.com/en/articles/integer-scaling/

- [ ] Fix "shimmer" aka "wave" effect with moving images when `PAR != 1.0`.
- [ ] Add API to change PAR at runtime.


dbalsom:
Sorry to be a nag but what is the state of this PR? Anything we can do to help?

parasyte:
There aren't any updates for this PR other than what is already here. Specifically, the extra context provided in #262 kind of obsoletes the implementation in this PR. In short, this does allow for non-square pixel aspect ratios, but it comes at the cost of "pixel shimmering" in animations, especially when things move horizontally. Texture filtering like the two-pass implementation in #262 can address that problem.

There is still an unresolved issue there; how to integrate the second pass into the library as a first-class feature (that PR currently requires it to be added to a project as a custom renderer). If you are interested in helping drive the effort, that would be the place to start.

parasyte:
This will need to be redone using the same technique introduced in #425. In fact, I believe it is just an additional parameter needed to fully implement non-square pixel aspect ratios in the `scale_fill.wsgl` shader that was introduced in that PR.

PR #153, opened 2021-03-29
rename resize_surface to resize

BarePotato:


parasyte:
`resize_surface` is intentional. If you want to run the winit example with an older version of the `pixels` crate, you can check out the example with a tag matching the version. E.g. https://github.com/parasyte/pixels/tree/0.2.0/examples/minimal-winit

parasyte:
The rename happened in #149 and this PR breaks CI.

PR #172, opened 2021-06-06
Rewrite shaders in WGSL

### LINKED ISSUE #141: Port shaders to WGSL
This is a followup to https://github.com/parasyte/pixels/pull/134#issuecomment-780035330

I have an experimental Work In Progress that uses a slightly older version of `naga` to compile the WGSL. It needs to be updated with the latest changes contributed by the community, which includes `wgpu` 0.7. I haven't published a branch for it, yet. It needs some cleanup and updates.

The experimental WIP uses `wgpu` 0.6, so I had to compile WGSL to SPIR-V before passing the shader to `wgpu`. 0.7 is able to accept WGSL directly without the intermediate compilation step. There may also have been some syntax changes to WGSL since I last worked on it.

Finally, only the shader for the scaling renderer has been ported. The custom shader example still needs work.

parasyte:
Fixes #141


JMS55:
Btw next wgpu will have an include_wgsl macro.

PR #187, opened 2021-08-31
Update to wgpu 0.10

parasyte:
- It would be nice to return an error from the render function
- `imgui-winit` is still a WIP (open PR: https://github.com/Yatekii/imgui-wgpu-rs/pull/66)
- `minimal-fltk` flashes wildly! :(

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ pollster = "0.2"
 raw-window-handle = "0.3"
 thiserror = "1.0"
 ultraviolet = "0.8"
-wgpu = "0.9"
+wgpu = "0.10"
 
 [dev-dependencies]
 pixels-mocks = { path = "internals/pixels-mocks" }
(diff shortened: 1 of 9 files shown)


parasyte:
Just so I don't forget: I need to verify if `pixels.surface_size` is still necessary. The new `wgpu::Surface` basically has the same information as our `Surface` struct (IIRC). We might be able to trim some fat!

MoAlyousef:
Hi
Thanks for this. Regarding the fltk example, can you try replacing the:
```rust
        win.redraw();

        // Calls to redraw in the event loop require an explicit sleep
        thread::sleep(Duration::from_millis(SLEEP));
```
part of the loop with:
```rust
        app::flush();
        app::awake();
```
This seems to resolve the issue of flashing for me on Windows.

parasyte:
Thanks for the suggestion! I'll give this a shot tomorrow when I get back to resolving the other remaining issues.

parasyte:
Just converted a project over to this branch and it was a drop-in replacement. I think this is good to go.

PR #190, opened 2021-09-02
Properly clip pixel buffer when larger than surface texture

### LINKED ISSUE #186: Larger resolutions fail to render
I've been trying to use Pixels in a personal project, but I've hit what seems to be a bug with resolution handling. It seems that if you have a larger winit window, therefore a large surface texture, the first attempt at calling `Pixels.render()` crashes with a wgpu validation error. I think its related to the aspect ratio it ends up using,
```
[2021-08-26T01:52:49Z ERROR wgpu::backend::direct] wgpu error: Validation Error

    Caused by:
        In a RenderPass
          note: encoder = `<CommandBuffer-(0, 1, Vulkan)>`
        In a set_scissor_rect command
        Invalid ScissorRect parameters
```

I dug through just about the entire issue tracker but I wasn't able to find anything related. The closest thing seems to be [this open PR](https://github.com/parasyte/pixels/pull/151) about aspect ratios. Interesting, the minimal example on this branch doesn't panic rendering at 2K. I'm not sure how intentional this is though or if there's something else needed for the "ideal" result. 

Steps to reproduce:
1. `git clone https://github.com/parasyte/pixels.git`
2. `cd pixels/examples/minimal-winit`
3. `sed -i "s/320/2560/; s/240/1440/" src/main.rs`
4. `cargo run --release`

OS: Windows 10 or Linux (X11 vs Wayland doesn't matter)

parasyte:
- Fixes #186

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -464,13 +464,11 @@ impl Pixels {
         );
 
         let pos = self.scaling_matrix_inverse * pos;
+        let offset_width = pixels_width.min(physical_width) / 2.0;
+        let offset_height = pixels_height.min(physical_height) / 2.0;
 
-        let pos = (
-            pos.x / pos.w + pixels_width / 2.0,
-            -pos.y / pos.w + pixels_height / 2.0,
-        );
-        let pixel_x = pos.0.floor() as isize;
-        let pixel_y = pos.1.floor() as isize;
+        let pixel_x = (pos.x / pos.w + offset_width).floor() as isize;
+        let pixel_y = (-pos.y / pos.w + offset_height).floor() as isize;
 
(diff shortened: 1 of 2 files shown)


parasyte:
@BlackHoleFox, please try this branch. If it works for you, I'll consider that a positive review of the PR. :P

BlackHoleFox:
@parasyte I swapped in this branch and it was pretty much a drop-in replacement. The results are great too :) No `wgpu` errors occur and the top bit of the image is no longer missing either, which makes a large improvement over the previous branch I was trying out.

parasyte:
Thanks for confirmation! I'll go with this PR as-is and create a new release soon.

PR #204, opened 2021-09-17
Use crab emoji instead of markdown

PrototypeNM1:
Github style markdown emoji don't render on crates.io. Use standard emoji instead.


parasyte (approved):
Thanks!

PR #218, opened 2021-10-28
Initial WebGL2 support.

### LINKED ISSUE #115: WASM support
It would be awesome if we could get this crate working with WASM! I've looked into this a little bit, and here's what I've found so far. I figure this is more of a tracking issue than anything else, because it looks like the implementation is blocked on more upstream support in wgpu.

First (and easiest), we need to expose an API to create a `Pixels` instance without blocking. Blocking is not allowed in WASM. This is pretty easy: just add an `async` version of `build()` that doesn't use `pollster`.

After doing this, I modified the `minimal-winit` example to add a canvas and spawn the (now async) example in WASM. I also modified the default texture format to be `Rgba8Unorm` because SRGB is not currently supported in Chrome.

Then, I built the example while targeting `wasm32-unknown-unknown`, and generated the bindings with `wasm-bindgen-cli`

```
$ RUSTFLAGS=--cfg=web_sys_unstable_apis cargo build --manifest-path examples\minimal-winit\Cargo.toml --target wasm32-unknown-unknown
$ wasm-bindgen --out-dir target\generated --web .\target\wasm32-unknown-unknown\debug\minimal-winit.wasm
$ python3 -m http.server
```

Next, I started an HTTP server in the root directory, and loaded the server in my browser (Chrome canary). Further instructions on how to do this [here](https://github.com/gpuweb/gpuweb/wiki/Implementation-Status).

Here's where I'm stuck. I get a panic because `buffer_get_mapped_range_mut` (called by `create_buffer_init`) isn't implemented for the web backend in wgpu:

```
minimal-winit.js:333 panicked at 'not implemented', C:\Users\aruss\.cargo\registry\src\github.com-1ecc6299db9ec823\wgpu-0.6.0\src\backend\web.rs:1157:9

Stack:

Error
    at imports.wbg.__wbg_new_59cb74e423758ede (http://localhost:8000/target/generated/minimal-winit.js:339:19)
    at console_error_panic_hook::Error::new::h4bc2c46dd82b9ea7 (http://localhost:8000/minimal-winit_bg.wasm:wasm-function[3405]:0xa6966)
    at console_error_panic_hook::hook_impl::h810a5a414ea0321f (http://localhost:8000/minimal-winit_bg.wasm:wasm-function[390]:0x3e1b2)
    at console_error_panic_hook::hook::h47f770812af45fc2 (http://localhost:8000/minimal-winit_bg.wasm:wasm-function[3935]:0xadd3a)
    at core::ops::function::Fn::call::heaa8aaab91c5f390 (http://localhost:8000/minimal-winit_bg.wasm:wasm-function[3043]:0xa1033)
    at std::panicking::rust_panic_with_hook::h3e6619a8809a443a (http://localhost:8000/minimal-winit_bg.wasm:wasm-function[896]:0x6406f)
    at std::panicking::begin_panic::hee06aa61d0ace97b (http://localhost:8000/minimal-winit_bg.wasm:wasm-function[1504]:0x7e321)
    at <wgpu::backend::web::Context as wgpu::Context>::buffer_get_mapped_range_mut::h2675a7cb5a19d270 (http://localhost:8000/minimal-winit_bg.wasm:wasm-function[2536]:0x97f65)
    at <wgpu::Device as wgpu::util::DeviceExt>::create_buffer_init::ha62d9f473e3f1ad2 (http://localhost:8000/minimal-winit_bg.wasm:wasm-function[151]:0x17f47)
    at pixels::renderers::ScalingRenderer::new::hedee3d7f7c822c12 (http://localhost:8000/minimal-winit_bg.wasm:wasm-function[130]:0xcb89)
```

[You can see the code that I wrote to do this on my fork](https://github.com/euclio/pixels/tree/wasm).

parasyte:
- Closes #115


parasyte:
Yep, thought about that very briefly. I don't want winit as a direct dependency, but we can absolutely wrap up the boilerplate into its own crate. As long as it can take care of things like resize, I believe that would be a huge improvement over rolling your own WASM container.

MarkAnthonyM:
I think I got a version of this example going that has some resize capability, but it's kinda janky. It doesn't smoothly resize, but jumps to bigger and bigger sizes based on the current browser window size.
like so: https://media.giphy.com/media/5FPB18f3h6oQRt7pWE/giphy.gif

I honestly wasn't too sure what I was doing, just used other code I found as a template and sort of fiddled with it to get it going. So I'm not sure if I can figure how to get it resizing smoothly, but I'll still give it a shot. If I get it working, is there a way to include the changes in this specific pull request?

parasyte:
@MarkAnthonyM You can create a PR against the [`webgl2-support` branch](https://github.com/parasyte/pixels/tree/webgl2-support) while this PR is open. I'll gladly accept any improvements here.

FWIW, I looked briefly at how the `egui` demo handles resize for web builds. It just does [its own DOM access to determine the browser's client size](https://github.com/emilk/egui/blob/78dfde40b2fbda8d44947b8ebf186eb42568a400/egui_web/src/lib.rs#L192-L229). I'm not sure if this is the right way to do this, since `winit` appears to expect it has exclusive control over the canvas size according to [this comment](https://github.com/rust-windowing/winit/issues/1491#issuecomment-696044564).

It may or may not be useful, but it's an extra datapoint and some prior art in the area.

JMS55:
Resize support would be great! Is the non-smooth resizing not just pixels only snapping to integer multiples? I.e., pixels won't scale your game by 1.343 or anything, only 1x, 2x, 4x, etc...

parasyte:
The linked gif looks correct, to me. 🤷

MarkAnthonyM:
@JMS55 Oh ok, I didn't know about that

I was under the impression that the height and width of pixel's rendering area was suppose to grow at the same height and width values that the browser client does. Like following closely to the edges of the browser window, leaving no black borders. Now that I think about it a little bit, I guess that wouldn't really make sense from pixel's perspective.

In that case, if this is the expected behavior, I'll just clean up the code and send off a PR.

parasyte:
> I was under the impression that the height and width of pixel's rendering area was suppose to grow at the same height and width values that the browser client does. Like following closely to the edges of the browser window, leaving no black borders.

The expected behavior can be accomplished on the caller-side. For instance, the `minimal-imgui` example resizes the pixel buffer to match the window size (other examples do not): https://github.com/parasyte/pixels/blob/eb07e61f5c25a972a422d6face10fb5880267111/examples/imgui-winit/src/main.rs#L102-L113

> In that case, if this is the expected behavior, I'll just clean up the code and send off a PR.

That sounds good to me! Thank you.

MarkAnthonyM:
> With this extension trait, the WASM-specific part of `run` becomes just a call to `window.init_web()`. I would really like to encapsulate the bifurcated `main`, though. The best I've come up with for that is just moving it to a bare function in whatever crate this boilerplate goes into:
> 
> ```rust
> fn main() {
>     pixels_web::app(run());
> }
> ```
> 
> For general purpose, the `init_web` method should panic if a canvas was already added to the body. Prevent calling it twice and doubling the canvas element and resize listener.
> 
> I'm open to bikeshedding the method names and the design. Any ideas for improving `main`?

I think your idea for abstracting away the fluff currently in main makes sense. I can't think of anything else other then decorating stuff with macros, but that seems a little overkill. I did a quick dry run, as I wasn't sure what the signature of `pixels_web::app()` should look like, and it worked out fine. But I started wondering about where pixels_web would be located. Would it be top level with pixels? Something like?
```
root
|------pixels
|       |------examples
|       |------internals
|       |------src
|------pixels_web
        |------src
```

parasyte:
> I did a quick dry run, as I wasn't sure what the signature of `pixels_web::app()` should look like, and it worked out fine.

Me too. I came up with:

```rust
pub fn app(future: impl Future<Output = ()> + 'static)
```

> But I started wondering about where pixels_web would be located. Would it be top level with pixels?

After I put my thoughts into usable code, I recognized that it doesn't have any relation to `pixels` at all. It's just an extension trait for `winit` and an abstraction for polling a future in WASM and native environments. I would almost want to call the crate `winit-web` because of that! It doesn't seem to matter where it lives. Either here (a top-level crate as you've shown) or in another repo by another owner entirely... I am hesitant to reorganize this repo just for a new top-level crate that only indirectly has anything to do with the main library.

You can create a new repo for it if you like! Or if you would rather have someone else maintain it, I can.

MarkAnthonyM:
Yea, I'll throw the crate together as a separate repo and maintain it. [At least until winit figures out how exactly it wants to handle the situation related to the resizing boilerplate.](https://github.com/a-b-street/abstreet/pull/388#issuecomment-736530257) That leaves me wondering about main again though...

Anyway, I'll send another PR to update the example when I get everything going

parasyte:
I read in one of the winit issues that it will also need a similar workaround to receive mouse and touch events. `pixels` has a few methods for converting between window coordinates and pixel buffer coordinates, which is intended to be used with pointer position events. The `Conway` example uses it for demonstration purposes. That would be a nice bonus for the crate to support. And I’m sure it could grow into something of a stop-gap for several web events that winit doesn’t support yet.

parasyte:
I think this is pretty much done from the `Pixels` API standpoint. The rest of the work here is making the `minimal-web` example smaller by abstracting some of the special sauce into a new crate, as discussed above. But I don't think that should block us. This PR will enable `pixels` on the web and is plenty enough reason for a new release ASAP.

 Any thoughts from collaborators?

parasyte:
Since no one has responded over the weekend, I'll take that as good-to-go. 🙂

PR #223, opened 2021-11-12
Set default device limits to `downlevel_webgl2_defaults`

parasyte:
- Fixes #139
- See also #115

PR #226, opened 2021-11-14
Use adapter limits (unless overridden)

parasyte:
- Fixes a regression caused by #223
- `downlevel_webgl2_defaults` only supports 2048x2048 pixels buffers
- This will allow pixel buffers up to the max supported by the hardware

TODO:

- [x] Test with Windows 10/Vulkan
- [x]  Test with macOS/Metal
- [x] Test with Linux/Vulkan
- [x] Test with WebGL2
- [x] Test with Raspberry Pi 4


parasyte:
Can't really test on Linux/Vulkan, here. But Linux/GL is working (requires `LIBGL_ALWAYS_SOFTWARE=1` and `WGPU_BACKEND=gl` environment variables).

parasyte:
Pi4 works! I'm pretty confident this will not cause any problems on Linux/Vulkan with other hardware/drivers. (Pi4 is using Vulkan in my test.)

PR #228, opened 2021-11-20
Simplify vertex attributes

parasyte:
- Vertex UVs can be computed from the positions, which saves a small
  amount of code and a small amount of bandwidth (both inconsequential).
- This is mostly for readability.

### DIFF
--- a/examples/custom-shader/shaders/noise.wgsl
+++ b/examples/custom-shader/shaders/noise.wgsl
@@ -8,10 +8,9 @@ struct VertexOutput {
 [[stage(vertex)]]
 fn vs_main(
     [[location(0)]] position: vec2<f32>,
-    [[location(1)]] tex_coord: vec2<f32>,
 ) -> VertexOutput {
     var out: VertexOutput;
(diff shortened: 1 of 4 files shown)


parasyte:
Hmm, this PR made me realize that the texture coordinates on the custom shader example were flipped vertically. The pixel buffer is upside down! It's hard to tell when it's just a moving square, but it's supposed to start near the upper-left corner and move to the lower-right.

The problem is that I was doing the texture coordinate correction in the transformation matrix instead of in the UVs themselves. Removing the vertical flip from the transformation matrix means that both shaders can do the same UV computation to correct the vertical axis. This fixes an accidental cancelation effect.

`wgpu` uses Metal/DirectX normalized coordinates (which can be [seen in the `wgpu` readme](https://github.com/gfx-rs/wgpu#coordinate-systems)) This means our triangle has its right-angle in the lower-left corner and UV `(0, 0)` starts half-way up the left side and goes down-right to `(1, 1)` half-way along the bottom of the triangle.

PR #230, opened 2021-12-01
Update dependencies

parasyte:
Update examples:

- [x] `conway`
- [x]  `custom-shader`
- [x] `imgui-winit`
- [x] `invaders`
- [x] `minimal-egui`
- [x] `minimal-fltk`
- [x] `minimal-sdl2`
- [x] `minimal-web`
- [x] `raqote-winit`


parasyte:
`egui` PR is here: https://github.com/emilk/egui/pull/930

`imgui` will be a lot more work, since they have an unusual collection of features to support several versions of `winit`.

`winit` 0.26.0 has at least one show-stopper bug which could hypothetically be released as a patch version like 0.26.1: https://github.com/rust-windowing/winit/pull/2078 The expectation is that this will be required before Rust 1.58.0 is released around mid-January.

parasyte:
The breaking change for macOS linking was reverted on the next stable, so we should be good here for a while.

`winit` 0.26 support landed for `imgui-winit-support` in https://github.com/imgui-rs/imgui-rs/pull/583

parasyte:
`minimal-egui` is blocked on https://github.com/hasenbanck/egui_wgpu_backend/pull/51

Issue #80, opened 2020-05-22
Fix `pixels-mocks` crate


parasyte:
We have cargo doctests enabled in CI, but many of them do not run because they are tagged with `no_run`. The reason for this is the `pixels-mocks` crate doesn't generate valid code (it just compiles).

This has resulted in at least one bad example that would have hit an assertion failure: https://github.com/parasyte/pixels/blob/de21862c7740623cdaa9c0836b25295976151710/src/lib.rs#L352-L367

The last assertion should compare against `(11, 239)`

Fixing the `pixels-mocks` crate enough that the examples can run would have prevented this error.

PR #250, opened 2022-01-08
Fix invalid texture sizes

### LINKED ISSUE #240: Properly handle dubious texture sizes
There is a `wgpu` issue that we need to work around: https://github.com/gfx-rs/wgpu/issues/1010 I started experimenting with a patch that makes texture creation fallible by checking device limits and changing some APIs to return `Result`.

I ran into a number of problems with automated testing. The big one is that `pixels-mocks` is still not runnable (but it's ok for compilation tests). See #80. Using `winit` in integration tests has some issues with creating the event-loop off the main thread, at least on Windows. It can be done with platform-specific extensions. And then tests may not be able to run on all platforms.

parasyte:
- Makes methods fallible when they create textures.
- Breaking changes:
  - `Pixels::Error` is now marked `#[non_exhaustive]`
  - `Pixels::resize_buffer()` and `Pixels::resize_surface()` return `Result<_, TextureError>`.
  - `Pixels::resize_buffer()` no longer panics on invalid inputs.
- Correctly handle window resize in fltk example.
- TODO: tests
- Closes #240

I haven't written any tests for this, because I still don't have a great solution to the issue raised in #80.

This does add a lot of noise to the examples. I can perhaps address some of that with `Result` transformation methods. The "print error and exit event loop" pattern is very common in these examples but might need more intelligent handling in a real application.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,6 +1,5 @@
 use crate::renderers::{ScalingMatrix, ScalingRenderer};
-use crate::SurfaceSize;
-use crate::{Error, Pixels, PixelsContext, SurfaceTexture};
+use crate::{Error, Pixels, PixelsContext, SurfaceSize, SurfaceTexture, TextureError};
 use raw_window_handle::{HasRawDisplayHandle, HasRawWindowHandle};
 
 /// A builder to help create customized pixel buffers.
@@ -321,7 +320,7 @@ impl<'req, 'dev, 'win, W: HasRawWindowHandle + HasRawDisplayHandle>
                 render_texture_format,
                 clear_color,
                 blend_state,
-            );
+            )?;
 
         // Create the pixel buffer
         let mut pixels = Vec::with_capacity(pixels_buffer_size);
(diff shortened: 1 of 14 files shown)


src/builder.rs
```diff
@@ -347,6 +347,28 @@ impl<'req, 'dev, 'win, W: HasRawWindowHandle> PixelsBuilder<'req, 'dev, 'win, W>
...
+/// # Errors
+///
+/// - [`TextureError::TextureWidth`] when `width` is 0 or greater than GPU texture limits.
+/// - [`TextureError::TextureHeight`] when `height` is 0 or greater than GPU texture limits.
+pub fn check_texture_size(
+    device: &wgpu::Device,
+    width: u32,
+    height: u32,
+) -> Result<(), TextureError> {
+    let limits = device.limits();
```
parasyte, line 361: It will probably be better to store the limits in the constructor instead of making this call each time. I don't know what the exact overhead is, but we only need one of many limits here, and the limit is static.

Issue #249, opened 2022-01-06
Feature Request: option to decouple window surface texture format from `render_texture_format`


AnonymousDapper:
I'd like an option to use different texture formats for the ScalingRenderer output and the underlying SurfaceConfiguration for using external rendering pipelines that require "exotic" formats (such as `Rgba16Float` for HDR), while still using a hardware-supported format for the surface.

I believe this can be done in a matter of a few lines, by adding a new member (i.e. `surface_texture_format`) to Pixels to take the place of the current `render_texture_format`, while repurposing `render_texture_format`. (This may cause confusion, but I think it keeps the names where they make the most sense)

If this is a welcome change I'd be glad to open a PR or discuss implementation needs.

parasyte:
Yeah, that is something that can be done pretty easily! I haven't had a need for it myself, but I'll accept a PR.

PR #257, opened 2022-01-16
Decouple render format from surface format (#249)

AnonymousDapper:
This is possibly a breaking change, and I'm open to leaving `render_texture_format` as it is and finding a different name for the new value


parasyte:
This looks good. Thanks for the contribution!

PR #260, opened 2022-02-05
Make BlendState configurable

JMS55:
Makes the BlendState configurable so I can use the alpha channel as a flag for a custom shader.

I also reworded the documentation for clear_color a bit.

I checked and it works correctly, but let me know if I'm not threading the blend_state through each struct correctly.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -391,6 +391,7 @@ impl<'req, 'dev, 'win, W: HasRawWindowHandle> PixelsBuilder<'req, 'dev, 'win, W>
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 pub(crate) fn create_backing_texture(
     device: &wgpu::Device,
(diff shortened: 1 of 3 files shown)


src/builder.rs
```diff
@@ -189,11 +191,37 @@ impl<'req, 'dev, 'win, W: HasRawWindowHandle> PixelsBuilder<'req, 'dev, 'win, W>
         self
     }
 
-    /// Set the default clear color.
+    /// Set the blend state.
+    ///
+    /// Allows customization of how to mix the new and new existing pixels in a texture
```
parasyte, line 196: Looks like a small typo here.

```suggestion
    /// Allows customization of how to mix the new and existing pixels in a texture
```

parasyte (approved):
Thank you! This looks good to me. I have just one nitpick in the docs.

Issue #106, opened 2020-07-20
Black border on resizing


FloVanGH:
When I resize the window on the minimal-winit example on macOS Catalina I've got a black border around the content. I have the same problem with the other examples.

![image](https://user-images.githubusercontent.com/6715107/87903318-fd604180-ca5b-11ea-85d6-06c1afaf01f1.png)

JMS55:
This is intended. Pixels will only scale the game when it can scale to a perfect integer, e.g. 2x, 3x, 4x, etc the initial size. Otherwise, the black borders are added.

parasyte:
I'm going to close this, since it is working as intended. However, I will accept feature requests for configuring how the scaling works. E.g. #89

Issue #89, opened 2020-06-14
Feature request: force a particular scaling factor


euclio:
I recently ported my emulator to use pixels, and it's great! I love the automatic scaling to the window size.

However, I would also like to allow my users to select a scaling factor that they prefer, regardless of window size. I tried passing my own values to `pixels.resize`, but this doesn't do what I want.

parasyte:
`pixels` doesn't integrate directly with any windowing provider (like `winit` or `sdl2`), If you want to resize the window, you need to use the API supported by your window provider. After resizing the window, you can call the `resize` method to fit the output texture to the window. (This is usually going to be called in response to resizing the window as part of application-specific integration with a window provider.)

Does that help? I don't think there is anything more `pixels` can do to resize your window.

euclio:
Sorry, let me clarify. To use your example of a 160x120 pixel buffer, I would like the ability to force the scale of the buffer display to 1x or 2x, even if the window itself is much larger (say, 640x480) and could support a scale of 4x.

parasyte:
Ok. It would just be a fixed scaling factor instead of dynamic (based on window inner size). This can be done.

PR #262, opened 2022-02-09
Add a second renderer for scaling to non-integer sizes

parasyte:
Related issues:

- #106
- TBD whether this should address #89 

The current state of this PR adds an example which implements a custom renderer for doing two-pass scaling to maintain high quality with the sharpest pixels possible on arbitrary image sizes. I would like to evolve this a bit to pull this custom renderer into the library and make it customizable.

There are a few gotchas to consider:

1. `Pixels` maintains an inverse scaling matrix for translating pointer coordinates to pixel buffer coordinates and back. Using a two-pass renderer needs to be aware of this.
    - Let `A` be the scaling matrix for the first pass and `B` be the scaling matrix for the second pass.
    - Right now, the inverse scaling matrix is always <code>A<sup>-1</sup></code>
    - With two passes, it will need to be changed to <code>(AB)<sup>-1</sup></code>
2. The naive approach that I originally wanted to take for implementing this second pass is to just make the user do it with `Pixels::render_with()` and our `PixelsContext` would allow referencing both renderers. I now see this is error prone. Not only because the inverse matrix will complicate this, but also because it would be possible to accidentally swap the two passes and end up with a bad result.
3. The `ScalingRenderer` is recreated every time the pixel buffer is resized. This has been fine up to now, but it complicates things with multiple passes. I think it makes more sense to make the renderer mutable and design a trait that can generically update the renderer when properties like the pixel buffer size changes.

For these reasons, I think a better API would be allowing some configuration at the `PixelsBuilder` level to choose whether you want one-pass or two-pass scaling, plus any scaling options. And those options can all follow through to correctly initialize and update the inverse scaling matrix and renderer.

It would be nice if `PixelsContext` could own a trait object for the renderer. Instead of owning two separate renderers, it would just own one (and the two-pass renderer would own the first pass). This would make swapping renderers easy in the future and could also allow changing the renderer at runtime (if that's a thing we want to do).

----

This _might_ be YAGNI, but given that we have at least the one issue regarding the border and the fact that it is not at all obvious why the border exists, I think dealing with this head-on is appropriate.


parasyte:
I was thinking about this the other day, but I'll log it here for documentation: I think this could be a suitable replacement for the outstanding issue in #151. In short, that issue is enabling horizontal texture filtering (smoothing) without introducing a second render pass. Since we need a second render pass _anyway_ for this feature, I think it is worth taking advantage of that in #151 accordingly.

I'll have to revisit that PR after I sort out the outstanding issues in this one. :)

parasyte:
There are better ways to do this. #425 added scaling modes that produce minimal aliasing artifacts. That addresses the non-integer sizes part of this PR. The same techniques can be extended to support non-square pixel aspect ratios (the other part of this PR).

PR #288, opened 2022-07-07
Allow accessing frame on immutable buffer

squ1dd13:
I have a trait like the following:
```rs
pub trait Buffer {
    /// Returns the size of the buffer.
    fn size(&self) -> Size;

    /// Returns a copy of the pixel data for the given coordinates, or `None` if no such pixel
    /// exists.
    fn pixel_chunk(&self, x: u32, y: u32) -> Option<[u8; 4]>;

    /// Returns a mutable reference to the pixel data for the given coordinates, or `None` if no
    /// such pixel exists.
    fn pixel_chunk_mut(&mut self, x: u32, y: u32) -> Option<&mut [u8]>;

    // -snip-
}
```
I do not know what users of the trait will do with the values returned from `pixel_chunk`, but any implementation of this trait that is built on top of a `Pixels` instance will not be able to implement `pixel_chunk` without having some sort of interior mutability in place because `Pixels::get_frame` only takes `&mut self` because it returns a mutable reference.

This pull request adds another method, `get_frame_immut`, that returns an immutable reference to the underlying frame, so that users can access the pixel data without needing a mutable reference to the `Pixels` structure that holds it.

For naming, I would have preferred that the original method was `get_frame_mut` and the new one just `get_frame`, but I've instead used `get_frame_immut` because `get_frame` is such an integral method to using this library, so renaming it would break a lot of code (and cause confusion because of the different mutability on the new method's return value).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -497,8 +497,8 @@ impl Pixels {
         &mut self.pixels
     }
 
-    /// Get an immutable byte slice for the pixel buffer. The buffer is _not_ cleared for you; it will
-    /// retain the previous frame's contents until you clear it yourself.
+    /// Get an immutable byte slice for the pixel buffer. The buffer is _not_ cleared for you; it
+    /// will retain the previous frame's contents until you clear it yourself.
     pub fn get_frame(&self) -> &[u8] {
         &self.pixels
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -497,6 +497,12 @@ impl Pixels {
         &mut self.pixels
     }
 
+    /// Get an immutable byte slice for the pixel buffer. The buffer is _not_ cleared for you; it will
+    /// retain the previous frame's contents until you clear it yourself.
+    pub fn get_frame_immut(&self) -> &[u8] {
+        &self.pixels
+    }
+
     /// Calculate the pixel location from a physical location on the window,
     /// dealing with window resizing, scaling, and margins. Takes a physical
     /// position (x, y) within the window, and returns a pixel position (x, y).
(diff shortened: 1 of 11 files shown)


squ1dd13:
Thanks for the feedback. I've updated the naming to `get_frame` and `get_frame_mut`.

parasyte:
There are a lot of examples that need to be updated with the new name. You can use `cargo test —all` to run tests locally. I’ll enable CI for this PR as well. Just be aware that it will send you notices when there are any failures.

squ1dd13:
I've updated all the stuff in the `examples` folder. `cargo text --all` succeeds on my end now.

src/lib.rs
```diff
@@ -493,10 +493,16 @@ impl Pixels {
 
     /// Get a mutable byte slice for the pixel buffer. The buffer is _not_ cleared for you; it will
     /// retain the previous frame's contents until you clear it yourself.
-    pub fn get_frame(&mut self) -> &mut [u8] {
+    pub fn get_frame_mut(&mut self) -> &mut [u8] {
         &mut self.pixels
     }
 
+    /// Get an immutable byte slice for the pixel buffer. The buffer is _not_ cleared for you; it
+    /// will retain the previous frame's contents until you clear it yourself.
```
parasyte, line 501: Let's remove the note about clearing the buffer from this one. The slice is immutable so there is no way to clear the buffer through it.

I would also go as far to say that the doc comment should specify more of its purpose. It seems to me the only point of this is to treat the pixel buffer as the source of truth for what will be rendered. While I never really envisioned it being used that way, I can see the merits. Sort of. I personally don't have any reason to want to read from the buffer, but maybe someone out there does. Blending in software or something?

squ1dd13: Blending precisely – I'm using this for compositing. I clear the buffer at the start of each frame, so at any point during rendering the pixel buffer represents what has been rendered below the current thing that is being rendered.

I'll update the comment to reflect this.

parasyte:
Thanks for your contribution! I don't know when the next release will be, but I do have some work to do on the crate to upgrade dependencies. It should be reasonably soon!

PR #289, opened 2022-07-18
Dependency updates (wgpu 0.13)

JMS55:
Initial work towards wgpu 0.13.

builder.rs::get_texture_format_size() needs updating.
So does the egui example.

I'm also not sure if we're handling the surface_texture_format correctly or not.

Git seems to have gotten confused - I guess that's what I get for forgetting to rebase my fork :/. I need to fix this.


parasyte:
`egui-wgpu` is has been updated to 0.13. We'll have to use a `git` path until it is released. So, the demo should not be much trouble to update!

The texture format size can be daunting, but the math tends to be pretty easy. I usually just read through the type docs to figure out what's new or changed, and cross-reference with wikipedia or some other doc sources for the compressed formats.

> I'm also not sure if we're handling the surface_texture_format correctly or not.

Do you have any more info on this? What might need correcting here?

Zageron:
Hey there, any update on a possible resolution to this? Anything I can do?
Thanks for all the hard work!

JMS55:
Unfortunately I've gotten quite busy, and do not know when I'll have time to finish this PR. Feel free to pick up where I left off. Iirc there's not a ton left, just the stuff I talked about in the first comment.

parasyte:
This work will be continued in #300 with a clean rebased history. Thank you!

PR #300, opened 2022-08-17
Update wgpu to 0.13

parasyte:
This rebases #289 on top of the current `main`. Thanks to @JMS55 for contributing the initial work!


JMS55:
This should probably bump bytemuck to 1.12 instead of 1.10. You may want to consider setting up dependabot. 

Otherwise lgtm.

Issue #276, opened 2022-06-02
Add a feature flag for webgl backend


KaminariOS:
Currently, if I want to use wasm-pack to build for wasm32, I have to specify 

```
[target.'cfg(target_arch = "wasm32")'.dependencies]
wgpu = { features = ["webgl"]}
```

The Cargo warns that `warning: dependency (wgpu) specified without providing a local path, Git repository, or version to use. This will be considered an error in future versions
`

To solve this problem, we should add 

```
[features]
web = ["wgpu/webgl"]
```
to the Cargo.toml of this crate.

parasyte:
You can add a dependency on `wgpu` in your crate, even if it is otherwise unreferenced. I'm not opposed to adding features, though.

KaminariOS:
A simple and safe way to do this is to add one line here:

```
[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
pollster = "0.2"
wgpu = { version = "0.12", features = ["webgl"]}
```

parasyte:
This was fixed by #320

PR #312, opened 2022-10-15
chore(deps): upgrade wgpu from 13 to 14

Zageron:


Zageron:
Green:
```
cargo build
cargo build --release
cargo check --workspace
cargo test --workspace
cargo run-wasm --release minimal-web
```

Zageron:
Edge case: Most examples didn't play well, because their dependencies aren't up to date.

src/builder.rs
```diff
@@ -488,7 +493,8 @@ const fn get_texture_format_size(texture_format: wgpu::TextureFormat) -> f32 {
         | Rg8Snorm
         | Rg8Uint
         | Rg8Sint
-        | Rgb9e5Ufloat => 2.0, // 16.0 / 8.0
+        | Rgb9e5Ufloat
+        | Depth16Unorm => 2.0, // 16.0 / 8.0
```
Zageron, line 497: I have no idea if this is correct. "Moved" from https://github.com/parasyte/pixels/pull/312/files#diff-e4f794ca308aa712cbad31feb714df7cada1bbe453a30e232e45891571430263L514

src/builder.rs
```diff
@@ -325,6 +327,8 @@ impl<'req, 'dev, 'win, W: HasRawWindowHandle> PixelsBuilder<'req, 'dev, 'win, W>
         let mut pixels = Vec::with_capacity(pixels_buffer_size);
         pixels.resize_with(pixels_buffer_size, Default::default);
 
+        let alpha_mode = surface.get_supported_alpha_modes(&adapter)[0];
```
Zageron, line 330: Not sure if this is how you'd want to do it, but I found some examples of people doing this rather than just using "auto".

parasyte:
Thanks for this! I have been out of state on a road trip for the last week, and I will finally have some time to take a look. Expect some feedback soon!

Cargo.toml
```diff
@@ -20,22 +20,21 @@ include = [
...
 [dependencies]
-bytemuck = "1.12"
-raw-window-handle = "0.4"
-thiserror = "1.0"
-ultraviolet = "0.9"
-wgpu = "0.13"
+bytemuck = "1.12.1"
+raw-window-handle = "0.5.0"
+thiserror = "1.0.37"
+ultraviolet = "0.9.0"
```
parasyte, line 26: I would prefer to remove all of the patch numbers. They aren't necessary and the only time they could potentially matter is in the rare case that the current patch version you are using is yanked and a newer patch version is not released to fix whatever caused the yank.

I would even go as far to remove the minor numbers from any version that is `>= 1`. That's what I have been doing in other crates that I've worked on recently.

Cargo.toml
```diff
@@ -20,22 +20,21 @@ include = [
...
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+wgpu = { version = "0.14.0", features = ["webgl"] }
```
parasyte, line 29: This feature is already included in the web example: https://github.com/parasyte/pixels/pull/312/files#diff-9b657ee98202349de6df432957197f4efc6115c95e488ba75064bb24f076d6e3R24 Is it needed here, too?

See also: #276. We can do either-or, but I don't see any reason to do both.

Cargo.toml
```diff
@@ -20,22 +20,21 @@ include = [
...
-winit = "0.27"
+winit = "0.27.4"
 
 [workspace]
-members = [
-    "examples/*",
-    "internals/*",
-    "run-wasm",
-]
+members = ["examples/*", "internals/*", "run-wasm"]
```
parasyte, line 40: I would really prefer to undo this stylistic change.

Cargo.toml
```diff
@@ -1,10 +1,10 @@
 [package]
 name = "pixels"
 description = "A tiny library providing a GPU-powered pixel frame buffer."
-version = "0.9.0"
+version = "0.10.0"
```
parasyte, line 4: I usually don't bump the version number until I am ready for a release. But I also know that this can lead to confusion. Just FYI.

examples/conway/Cargo.toml
```diff
@@ -15,7 +15,7 @@ env_logger = "0.9"
 getrandom = "0.2"
 line_drawing = "1.0"
 log = "0.4"
-pixels = { path = "../.." }
+pixels = { git = "https://github.com/parasyte/pixels.git", rev = "9f7896cd2f1c80989e8f29ec6e3b6687183250a4" }
```
parasyte, line 18: This need to be reverted.

examples/imgui-winit/Cargo.toml
```diff
@@ -13,8 +13,10 @@ default = ["optimize"]
 env_logger = "0.9"
 imgui = "0.8"
 imgui-wgpu = "0.20"
-imgui-winit-support = { version = "0.8", default-features = false, features = ["winit-26"] }
+imgui-winit-support = { version = "0.8", default-features = false, features = [
+    "winit-26",
+] }
 log = "0.4"
-pixels = { path = "../.." }
+pixels = { git = "https://github.com/parasyte/pixels.git", rev = "9f7896cd2f1c80989e8f29ec6e3b6687183250a4" }
```
parasyte, line 20: This needs to be reverted.

examples/imgui-winit/Cargo.toml
```diff
@@ -13,8 +13,10 @@ default = ["optimize"]
 env_logger = "0.9"
 imgui = "0.8"
 imgui-wgpu = "0.20"
-imgui-winit-support = { version = "0.8", default-features = false, features = ["winit-26"] }
+imgui-winit-support = { version = "0.8", default-features = false, features = [
+    "winit-26",
+] }
```
parasyte, line 18: I would prefer to keep the original style here.

examples/minimal-egui/Cargo.toml
```diff
@@ -12,9 +12,11 @@ default = ["optimize"]
 [dependencies]
 egui = "0.19"
 egui-wgpu = "0.19"
-egui-winit = { version = "0.19", default-features = false, features = ["links"] }
+egui-winit = { version = "0.19", default-features = false, features = [
+    "links",
+] }
```
parasyte, line 17: I'd prefer the original style, again.

examples/minimal-egui/Cargo.toml
```diff
@@ -12,9 +12,11 @@ default = ["optimize"]
...
 env_logger = "0.9"
 log = "0.4"
-pixels = { path = "../.." }
+pixels = { git = "https://github.com/parasyte/pixels.git", rev = "9f7896cd2f1c80989e8f29ec6e3b6687183250a4" }
```
parasyte, line 20: This needs to be reverted.

examples/minimal-fltk/Cargo.toml
```diff
@@ -10,7 +10,11 @@ optimize = ["log/release_max_level_warn"]
...
-fltk = { version = "=1.3.13", features = ["raw-window-handle", "no-images", "no-pango"] }
+fltk = { version = "=1.3.13", features = [
+    "raw-window-handle",
+    "no-images",
+    "no-pango",
+] }
 env_logger = "0.9"
 log = "0.4"
-pixels = { path = "../.." }
+pixels = { git = "https://github.com/parasyte/pixels.git", rev = "9f7896cd2f1c80989e8f29ec6e3b6687183250a4" }
```
parasyte, line 20: This needs to be reverted.

examples/minimal-sdl2/Cargo.toml
```diff
@@ -13,6 +13,6 @@ default = ["optimize"]
 env_logger = "0.9"
 fermium = { version = "20022.0.0", features = ["raw-window-handle"] }
 log = "0.4"
-pixels = { path = "../.." }
+pixels = { git = "https://github.com/parasyte/pixels.git", rev = "9f7896cd2f1c80989e8f29ec6e3b6687183250a4" }
```
parasyte, line 16: This needs to be reverted.

examples/minimal-tao/Cargo.toml
```diff
@@ -12,5 +12,5 @@ default = ["optimize"]
 [dependencies]
 env_logger = "0.9"
 log = "0.4"
-pixels = { path = "../.." }
-tao = "0.12"
+pixels = { git = "https://github.com/parasyte/pixels.git", rev = "9f7896cd2f1c80989e8f29ec6e3b6687183250a4" }
```
parasyte, line 15: This needs to be reverted.

src/builder.rs
parasyte: This seems ok to me. The docs claim this method returns at least one element. So this should never panic.

src/builder.rs
```diff
@@ -597,6 +602,6 @@ const fn get_texture_format_size(texture_format: wgpu::TextureFormat) -> f32 {
         Astc { block: B12x10, channel: _ } => 7.5, // 12.0 * 10.0 / 16.0
 
         // 12x12 blocks, 16 bytes per block
-        Astc { block: B12x12, channel: _ } => 9.0, // 12.0 * 12.0 / 16.0
+        Astc { block: B12x12, channel: _ } => 9.0,
```
parasyte, line 605: Please revert this change. The comment is there so that we can replace the single arbitrary float with an equation that matches the comments. Const float arithmetic is not currently supported in Rust.

parasyte (commented):
Here's my initial feedback. Most of it is style nitpicks. I'll let CI run, though and we'll see if there is anything left to do!

Zageron:
Fyi, none of the examples with revs compile with the current changes because they have dependencies that conflict with wgpu or window handle.

Zageron:
Ayy, congrats on the upgrade to 13 and the release of 10. :D

parasyte:
Superseded by #320

PR #317, opened 2022-11-14
Add tiny-skia example

jeffreyrosenbluth:
Add an example following like raqote-winit for tiny-skia.


parasyte:
Hi, thanks for the contribution! There is one lint error from clippy. You can either push a new commit that fixes that or give me the go-ahead and I can do the cleanups.

parasyte:
Nice! Thank you again. I will follow up with a few minor changes in further PRs.

PR #340, opened 2023-02-26
Add `wgpu::Adapter` API to users

AndroGR:
"Closes" #339 

There's also the alternative of strictly exposing only `wgpu::AdapterInfo`. Which may be desired if we don't want to give any access to the adapter itself for whatever reason (Keep in mind with the commits as they are currently we give at most a mutable ref to the adapter, which may allow someone to modify things they shouldn't). I avoided this because why hide a simple adapter when the crate exposes other parts of `wgpu`? And at worst, let users decide if they want to get access to the adapter.

If the function names aren't good (eg. _wgpu_ shouldn't be there) or we need to wrap around `wgpu::Adapter` then lmk, I'll fix it.\
Verified to work under GNU/Linux as expected.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -284,12 +284,13 @@ impl Pixels {
     /// use pixels::Pixels;
     /// 
     /// let mut pixels = Pixels::new(320, 240, surface_texture)?;
-    /// let adapter = pixels.get_wgpu_adapter();
+    /// let adapter = pixels.adapter();
     /// // Do something with the adapter.
     /// ```
-    pub fn get_wgpu_adapter(&self) -> &wgpu::Adapter {
+    pub fn adapter(&self) -> &wgpu::Adapter {
         &self.adapter
     }
+
     /// Resize the pixel buffer and zero its contents.
     ///
     /// This does not resize the surface upon which the pixel buffer texture is rendered. Use
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -280,12 +280,15 @@ impl Pixels {
     ///
     /// The adapter can be used to retrieve runtime information about the host system
     /// or the WGPU backend.
-    /// ```no_run
-    /// use pixels::Pixels;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -275,6 +276,36 @@ impl Pixels {
...
+    /// or the WGPU backend.
+    /// ```no_run
+    /// use pixels::Pixels;
+    /// 
+    /// let mut pixels = Pixels::new(320, 240, surface_texture)?;
+    /// let adapter = pixels.get_wgpu_adapter();
+    /// // Do something with the adapter.
+    /// ```
+    pub fn get_wgpu_adapter_mut(&mut self) -> &mut wgpu::Adapter {
+        &mut self.adapter
```
parasyte, line 306: I don't think this is needed, since all `wgpu::Adapter` methods only take `&self`. And I suspect we don't want to allow things like `std::mem::swap()` to be used here.

parasyte (changes requested):
Hi @AndroGR, thanks for the PR.

I was hesitant on this one following the conversation in https://github.com/parasyte/pixels/discussions/339#discussioncomment-5115888 but I think what you have here is the right thing to do for now. I have just one change to request, and that's removing the `mut` variant of the method. I believe it is not needed.

I would also rename the method to `adapter` or `get_adapter`. My current preference is `adapter`, and I might even rename `get_frame` to `frame`, et al. for consistency.

AndroGR:
> I would also rename the method to adapter or get_adapter. (My current preference is adapter, and I am planning to rename get_frame to frame, et al. in another PR for consistency.)

Well, I am going to name it `adapter` and I'll leave it up to you to decide for the style you prefer. Although since we are returning only a reference to it I imagine `adapter` makes more sense here (Like some popular crates do it)

parasyte (approved):
Thank you!

Issue #7, opened 2019-10-12
no_std simple-invaders


parasyte:
Lokathor proposed a tongue-in-cheek question about whether simple-invaders runs on N64. It could easily run on N64 if it was `no_std`! It can continue using `Vec` and friends from `core`, but will need to play nicely with a bump allocator. E.g. replacing temporary vectors with `arrayvec`, like this one:

https://github.com/parasyte/pixels/blob/cfab34f290998e38f6aeb059c3d938780acb349b/simple-invaders/src/lib.rs#L206

Another thing to do is moving `load_pcx` to a build script: https://github.com/parasyte/pixels/blob/cfab34f290998e38f6aeb059c3d938780acb349b/simple-invaders/src/loader.rs#L58-L114

And finally, replacing `rand_core::OsRng` with [`rand_xorshift`](https://lib.rs/crates/rand_xorshift) or [`randomize`](https://lib.rs/crates/randomize)

That should be it, for now! Whatever I decide to use for audio will have to be `no_std` compatible as well. My two options are some-DSP-crate or a decoder like [`lewton`](https://lib.rs/crates/lewton) or [`hound`](https://lib.rs/crates/hound) to create raw PCM samples in a build script.

PR #342, opened 2023-03-11
Make Rust standard library optional (WIP)

AndroGR:
Aims to close https://github.com/parasyte/pixels/issues/7, by removing all dependencies on the standard library. So far I've got it compiling but all examples fail to build, so there's some work to be done (I am mainly working on the `simple-invaders` part, so I will first fix that).

Milestones:
- [x] Switch away from `NonZeroU32` (Causes some unsafety by using raw integers, I'll see what can be done), `std`-exported types such as `Vec` and `Box`.
- [x] Add optional feature to enable the standard library (Enabled by default)
- [ ] Get `invaders` example working, optionally with better performance or reduced executable size.
- [ ] Find `std::collections::HashMap` replacement ([`hashbrown` exists](https://crates.io/crates/hashbrown), as well as `BTreeMap` available on the `alloc` crate)
- [ ] Stop depending on `thiserror` (It requires `std` for some reason, or so the compiler messages say: (There's [`anyhow`](https://crates.io/crates/anyhow), it will also do the job))
```sh
error[E0433]: failed to resolve: use of undeclared crate or module `std`
   --> src/lib.rs:127:10
    |
127 | #[derive(Error, Debug)]
    |          ^^^^^ use of undeclared crate or module `std`
```

**NOTE:**
`core::error::Error`, while available and working, requires nightly. This may mean the whole pull request will be blocked until it's stabilized (Or, we could just require nightly to disable the standard library).

### DIFF
--- a/examples/invaders/simple-invaders/src/collision.rs
+++ b/examples/invaders/simple-invaders/src/collision.rs
@@ -2,17 +2,17 @@
 
 use crate::geo::{Point, Rect};
 use crate::{Bullet, Invaders, Laser, Player, Shield, COLS, GRID, ROWS};
-use std::collections::HashSet;
+use alloc::collections::BTreeSet;
 
 /// Store information about collisions (for debug mode).
 #[derive(Debug, Default)]
 pub(crate) struct Collision {
-    pub(crate) bullet_details: HashSet<BulletDetail>,
-    pub(crate) laser_details: HashSet<LaserDetail>,
+    pub(crate) bullet_details: BTreeSet<BulletDetail>,
+    pub(crate) laser_details: BTreeSet<LaserDetail>,
 }
 
 /// Information regarding collisions between bullets and invaders, lasers, or shields.
-#[derive(Debug, Eq, Hash, PartialEq)]
+#[derive(Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
 pub(crate) enum BulletDetail {
     /// A grid position (col, row) for an invader.
     Invader(usize, usize),
@@ -23,7 +23,7 @@ pub(crate) enum BulletDetail {
 }
 
 /// Information regarding collisions between lasers and shields or the player.
-#[derive(Debug, Eq, Hash, PartialEq)]
+#[derive(Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
 pub(crate) enum LaserDetail {
     /// A shield index.
     Shield(usize),
--- a/examples/invaders/simple-invaders/src/geo.rs
+++ b/examples/invaders/simple-invaders/src/geo.rs
@@ -23,7 +23,7 @@ impl Point {
     }
 }
 
-impl std::ops::Add for Point {
+impl core::ops::Add for Point {
     type Output = Self;
 
     fn add(self, other: Self) -> Self {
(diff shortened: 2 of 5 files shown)


AndroGR:
Update 1: After some work it looks like unstable features are the only way to go with this. While searching for an alternative to `std::io::Cursor`, the only genuine crate that would still work similarly is the [`core_io`](https://crates.io/crates/core_io) crate which hasn't been updated in almost 2 years, causing compiler errors.

While considering all these, there's also the alternative of splitting the crate in two, writing different implementations for `#![no_std]` and normal environments. This would give the advantage of making the crate extremely robust and minimal, but with double the code and double the maintenance.

parasyte:
I don't think it is possible to make the main library `no_std` right now because `wgpu` depends heavily on `std`.

Making simple-invaders `no_std` should be the stopping point, since it can be used standalone without `pixels` or `wgpu`. I haven't been over the whole PR yet but I believe that's what you started with, which would be awesome to see completed! So thank you for your contribution.

I'll take a closer look at your work, but I like the idea behind the PR!

AndroGR:
It'd be pointless to have a single example running in `#![no_std]`, since any dependency that links to the standard library will use the standard library regardless. Obviously that means this PR is next to impossible to pull due to the dependencies.

The only thing we can be sure of is that `simple_invaders` should run without `std`. From there on it should probably be discussed further.

parasyte:
Yes. That is precisely why #7 is about the simple-invaders crate, and why the invaders example is split into two crates.

The original idea for that example was to showcase a game that can be used with `pixels` while its internals can also be used in other environments, including embedded systems.

There is some unwritten history, here. But I can provide some context if it will help explain the motivation. `pixels` originally came from an unrelated project that needed to simulate an NES screen. And when I needed to also simulate a screen for the Tiny Arcade I recognized the abstraction that was missing and rewrote the display part of that original project with `wgpu` (the original project itself used SDL2 at first and then was rewritten for the now-deprecated `gfx`).

Given that `pixels` is just an abstraction layer for putting some animated image into a window, it should be more clear that the `invaders` crate is also an abstraction layer that provides platform input, and `simple-invaders` is the platform agnostic layer that can conceptually run anywhere, even on a Nintendo 64 console. (Which I was also heavily working on at the time I released the first version of `pixels`.)

Other examples came out of necessity to highlight features of the crate. But you're right, simple-invaders is the only one that makes sense to remove the standard library. And that's by design.

AndroGR:
Judging by this, I will restart the PR working exclusively on the `simple_invaders` part since after all `pixels` looks like it needs a massive rewrite to remove the `std` dependency (And a pointless one, because `wgpu` still requires `std`).

AndroGR:
Should I make the entire example (Adding `#![no_std]` to `src/main.rs`, ...) or should I leave it here?

parasyte:
Not the entire example, only the `simple-invaders/src/lib.rs` needs the `#![no_std]` attribute.

AndroGR:
Then it's ready to go. I don't know why CI is failing though

parasyte:
CI is currently failing on lints because rustfmt finds differences. You can fix it by running `cargo fmt --workspace` and committing the changes.

AndroGR:
Fixed, ready to go!

parasyte (approved):
Nice, simple PR. Thank you for this!

PR #343, opened 2023-03-15
This commit changes SurfaceSize to public visibility.

rbermani:
It adds a an associated getter function for the Pixel struct to get the SurfaceSize.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,8 +50,8 @@ pub struct SurfaceTexture<'win, W: HasRawWindowHandle + HasRawDisplayHandle> {
 }
 
 /// A logical texture size for a window surface.
-#[derive(Debug)]
-struct SurfaceSize {
+#[derive(Debug, Clone)]
+pub struct SurfaceSize {
     width: u32,
     height: u32,
 }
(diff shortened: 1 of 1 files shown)


parasyte:
Hi. What's the use case for this? The reason it is currently not exposed is because the surface size is an internal detail. It is just a copy of the inner windows size.

And is it enough to return `&SurfaceSize` instead of a clone? The caller can clone it if they need an owner copy for some reason.

rbermani:
Hi, thank you for the context. I ran into a case where it was necessary to resize an existing surface, but only if desired dimensions differed from the existing surface. An alternative would be to store the state containing the current surface dimensions somewhere and check them before calling the surface resize function.

Rust defaults to returning by move, unless the type is Copy. So if clone() is not explicitly called, the function would try to move SurfaceSize, which would seem to be undesirable.

I don't mind just keeping this as a local change if it's deemed to have no utility.

parasyte:
> I ran into a case where it was necessary to resize an existing surface, but only if the dimensions differed from the existing surface.

Are you able to provide more details on this? I have never seen that case before. But to be fair, I mostly just use `winit` and react to the resize events. AFAIK, that's all that is really necessary for keeping the surface size in sync with the window. But perhaps I am missing something.

> Rust defaults to returning by move, unless the type is Copy. So if clone() is not explicitly called, the function would try to move SurfaceSize, which would seem to be undesirable.

Yes, but my question was about returning a borrowed struct, not owned copy of it. If you just need to access the struct members, they can also be made public and there is almost no need to have an owned copy. The struct can only be used internally by the library, anyway. Its members are what you are interested in.

PR #350, opened 2023-03-25
Added R key to reset game

imizao:
Hi guys, I added a [`reset_game`](https://github.com/imizao/pixels/blob/future-005/examples/invaders/simple-invaders/src/lib.rs#L463~L498) method to the example invaders. 

When the game is over, you can use the [`R`](https://github.com/imizao/pixels/blob/future-005/examples/invaders/src/main.rs#L173~L175) key to reset!

And use the [`min`](https://github.com/parasyte/pixels/pull/350/files#diff-da322482c505914fd24b6c98c55c60f8c5ffd8f3a7e2e43c7d53174d19d2591aL503~L521) function instead of the if function expression!

### DIFF
--- a/examples/invaders/simple-invaders/src/lib.rs
+++ b/examples/invaders/simple-invaders/src/lib.rs
@@ -564,6 +564,7 @@ impl Invaders {
         self.bounds.top_row = top;
         self.bounds.bottom_row = bottom;
 
+        // No more changes
         false
     }
 
--- a/examples/invaders/simple-invaders/src/lib.rs
+++ b/examples/invaders/simple-invaders/src/lib.rs
@@ -459,6 +459,43 @@ impl World {
             });
         }
     }
+
+    pub fn reset_game(&mut self) {
+        // Recreate the alien
+        self.invaders = Invaders {
+            grid: make_invader_grid(&self.assets),
+            stepper: Point::new(COLS - 1, 0),
+            direction: Direction::Right,
+            descend: false,
+            bounds: Bounds::default(),
+        };
+
+        // Empty laser
+        self.lasers.clear();
+
+        // Recreate the shield
+        self.shields = (0..4)
+            .map(|i| Shield {
+                sprite: Sprite::new(&self.assets, Frame::Shield1),
+                pos: Point::new(i * 45 + 32, 192),
+            })
+            .collect();
+
+        // Reset player position
+        self.player.pos = PLAYER_START;
+
+        // Remove bullet
+        self.bullet = None;
+
+        // Reset collision state
+        self.collision.clear();
+
+        // Reset game score
+        self._score = 0;
+
+        // Set gameover to false
+        self.gameover = false;
+    }
 }
 
 /// Create a default `World` with a static PRNG seed.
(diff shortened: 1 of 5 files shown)


examples/invaders/simple-invaders/src/lib.rs
```diff
@@ -459,6 +459,43 @@ impl World {
...
+        };
+
+        // Empty laser
+        self.lasers.clear();
+
+        // Recreate the shield
+        self.shields = (0..4)
+            .map(|i| Shield {
+                sprite: Sprite::new(&self.assets, Frame::Shield1),
+                pos: Point::new(i * 45 + 32, 192),
```
parasyte, line 480: We should probably move these magic numbers to `const` values at global (module) scope. If they need to be reused in multiple modules, we can create a `crate::consts` module for them.

examples/invaders/src/main.rs
```diff
@@ -165,6 +169,11 @@ fn main() -> Result<(), Error> {
                     return;
                 }
 
+                // Reset game
+                if g.game.input.key_pressed(VirtualKeyCode::R) {
```
parasyte, line 173: Please add a binding for controller inputs as well. Maybe the "select" button? I don't actually know what it's called any more on XBox/PS/Switch!

parasyte (changes requested):
Thank you! I love the refactor in `shrink_bounds()`.

IIRC what I wanted to do with this game was show a quick "game over" message and then automatically reset after a few seconds. It looks like you did the hard part implementing a reset method.

I'll accept this, but it would be a "more professional" looking game with the game over screen and not requiring the player to reset with a button.

imizao:
> Thank you! I love the refactor in `shrink_bounds()`.

Yes, I also like `shrink_bounds()`, but after the actual test, there are still problems _(such as: eliminating the entire line or sorting)_, so I removed it!

> IIRC what I wanted to do with this game was show a quick "game over" message and then automatically reset after a few seconds. It looks like you did the hard part implementing a reset method.
>
> I'll accept this, but it would be a "more professional" looking game with the game over screen and not requiring the player to reset with a button.

Yes, I agree with you!

Show game_over, adding a selection button interface will give users better interaction! This is what I want too!

For more professional games, you can also consider adding score and explosion effects...

But it is still a challenge for me now ; )

examples/invaders/src/main.rs
imizao: Maybe the "select" button? I don't know！LOL ; )

But I believe that if add the selected operation interface, this problem will be solved!

examples/invaders/simple-invaders/src/lib.rs
imizao: LGTM

parasyte:
> and explosion effects...

This was partially implemented a long time ago in #18! I was never quite satisfied with the way collision shapes are managed in that PR. This and some sound effects make the game really enjoyable.

imizao:
> > and explosion effects...
> 
> This was partially implemented a long time ago in #18! I was never quite satisfied with the way collision shapes are managed in that PR. This and some sound effects make the game really enjoyable.

Nice!

imizao:
image

What do I need to do with this?

parasyte:
There's a button to request reviews which will clear that status. I'll get to the review in the morning!

parasyte:
BTW here's that button I was talking about.

image

imizao:
Should I just click on it?

parasyte (approved):
This looks good! Are you planning to continue working toward the "game over" screen in this PR, or do you prefer this to be merged as it is? I'm ok even with the latter, but it seems kind of "half baked" to stand on its own.

I would just like a sense of what your plans are.

imizao:
Yes, I think I know what you mean!

Hmmm...my thought was to merge first!

Because I am still at work, I usually only consider dealing with open source work when I have enough time on weekends! In this process, I don’t want to have a PR hanging in my heart all the time.

Maybe I saw some parts that can be modified, or parts that can be optimized in the community, so I would consider participating in it! Just like the PR I submitted, I only want to do "Added R key to reset game" at the moment, maybe it may look like "half baked" at the moment! Don't worry about merging this PR! There may be another "I" who will appear to realize the "game over". Of course, if I achieve it, I will submit a PR without hesitation. I am very happy to contribute my meager strength to the community! Thank you also for your valuable advice, which gave me a lot of direction! I have learned a lot from cooperating with you, thank you very much!

parasyte:
Thank you for your contribution! It is objectively an improvement, even if it is not exactly what we want. :)

imizao:
Thank you for making a decision, things need to be perfected step by step, excellent products also need to come from constant polishing! Maybe someday I will find out that "Added R key to reset game" is wrong, and I will improve it too! Both we and our products need to undergo continuous self-improvement and self-iteration, so that we can grow slowly and become better slowly!  ; )

Issue #367, opened 2023-06-24
Android example


Randommist:
I think the project needs example code to run an android application. I've already managed to rewrite the `minimal-winit` example to make it work on Android. I used `winit` features `android-native-activity` and `cargo-apk` tool. In the future I might make a tetris game. Therefore, I want to discuss whether such an example is needed for `Pixels`

parasyte:
Hi! I don't have an Android device, but I would gladly accept a PR that introduces an example that can run on Android. It would only have to pass the minimum CI requirements.

Issue #238, opened 2021-12-06
API soundness issues


parasyte:
This was mentioned to me yesterday by @Lokathor. Trying to recreate a `Pixels` at runtime (e.g. in response to an event) causes validation errors and crashes.

I found an ugly workaround using `Option<Pixels>` and calling `take()` on it to drop the internal references before trying to recreate it. To me, this suggests the bug is related to https://github.com/gfx-rs/wgpu/issues/1463 (and this very long thread: https://github.com/rust-windowing/raw-window-handle/issues/71). The conversation is a bit hard to follow, but I think I can summarize it as: `RawWindowHandle` needs to be treated like a raw pointer that references the `HasRawWindowHandle` implementer. In this case, `Window` is the implementor, and `Surface` holds a `RawWindowHandle` pointing to it. This means that their lifetimes are required to be bound together, but there is nothing in the API contract enforcing that. It's left up to the caller of unsafe surface creation: https://github.com/parasyte/pixels/blob/4c6682e65bb417cdaa6e9964cbace5369149452d/src/builder.rs#L220

Whoops.

Pixels holds ownership of a surface, and it looks like having two surfaces alive pointing to the same `RawWindowHandle` is bad news bears. (This is just a best guess.)

Similarly, our API allows dropping the window after creating `Pixels`, and that is very clearly unsound. We can't borrow the window, since that would prevent moving it later (even though it's perfectly fine to do so). Almost every user moves their window into the event-loop closure to call `window.request_redraw()` on it, at a minimum.

I think the best way to solve this on our end is to take ownership of the window. That would prevent the issue as reported (can't recreate `Pixels` without creating a new `Window`) and it also solves this other related issue (can't drop `Window` without dropping `Pixels`). I don't know if shared ownership (with `Rc` or `Arc`) is acceptable, since that would probably just result in validation errors and crashes again if multiple surfaces are created from one window.

ZetaNumbers:
So we should mark `SurfaceTexture::new` as unsafe for now, right?

Lokathor:
That's a breaking change that's quick and dirty. If you're going to issue a breaking change you probably are better served to fix the whole problem in one go.

parasyte:
I do agree with the last comment. It seems like it should not be much more work to take ownership of the window. The advantage is that it would keep the API safe.

[...]

PR #368, opened 2023-06-25
Added example for android

Randommist:
solves the issue #367


examples/minimal-winit-android/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+                        SurfaceTexture::new(window_size.width, window_size.height, &_window);
+                    Pixels::new(WIDTH, HEIGHT, surface_texture).unwrap()
+                };
+                _window.request_redraw();
+                window = Some(_window);
+                pixels = Some(_pixels);
+            }
+            Event::Suspended => {
+                window = None;
+                pixels = None;
```
parasyte, line 44: Technically, `pixels` should be dropped before `window`.

The reason is subtle, but you can read more about it in #238. In short, `Pixels` owns a `SurfaceTexture` which borrows the `Window` [^1]. Therefore the `Window` should outlive the `Pixels`.

It's unlikely to cause problems in this case (since both are dropped in close proximity and they aren't shared by other threads or anything), so it's just a technicality.

[^1]: The normal way to declare this kind of dependency in Rust types is with a lifetime, but that would have been too restrictive. The linked ticket has more info, and some thoughts on how this API issue could be fixed.

examples/minimal-winit-android/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+            }
+            Event::RedrawRequested(_) => {
+                if let Some(_pixels) = &mut pixels {
+                    if let Some(_window) = &window {
+                        world.draw(_pixels.frame_mut());
+                        _pixels.render().unwrap();
+                        _window.request_redraw();
+                    }
+                }
```
parasyte, line 53: It's ok (and preferred) to shadow the `pixels` and `window` variables here. Also, the two `if-let` statements can be merged into one.

```suggestion
                if let (Some(pixels), Some(window)) = (&mut pixels, &window) {
                    world.draw(pixels.frame_mut());
                    pixels.render().unwrap();
                    window.request_redraw();
                }
```

examples/minimal-winit-android/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+            }
+            _ => {}
+        }
+        if let Some(_window) = &window {
```
parasyte, line 57: This should use `.is_some()` instead of pattern matching. I suspect you should see this one in the lint test results.

```suggestion
        if window.is_some() {
```

parasyte (changes requested):
Thank you for the contribution! It took me a few days to get around to this. I had to think a bit about drop order stuff, but revisiting the linked ticket, I'm confident that it does need to be addressed for soundness.

parasyte (approved):
Looks good! Thanks.

PR #373, opened 2023-08-09
Add present mode setters on Pixels

### LINKED ISSUE #372: No possibility of changing the surface `PresentMode` without recreating the whole `Pixels` instance
Pixels version: 0.13.0

I'm working on a project in which I would like to be able to toggle vsync on/off after my `Pixels` instance has been initialized. Currently, the only solution I found was to recreate my `Pixels` instance with this setting changed, using the `Option<Pixels>` workaround mentioned in #238 (it seems that problem in particular has been fixed on the main branch though)

Would it be possible for example to add some `enable_vsync` and `set_present_mode` methods on `Pixels`, similarly to what is currently exposed in `PixelsBuilder`. Those would set the present mode and reconfigure the wgpu surface. If that seems an ok proposal, I can submit a PR :slightly_smiling_face:

Kanabenki:
Closes #372.

Add `set_present_mode` and `enable_vsync` methods on `Pixels`. The documentation is almost identical to the corresponding methods on `PixelsBuilder` since they do the same thing but after the instance has been created.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -382,6 +382,31 @@ impl Pixels {
         Ok(())
     }
 
+    /// Enable or disable Vsync.
+    ///
+    /// Vsync is enabled by default. It cannot be disabled on Web targets.
+    ///
+    /// The `wgpu` present mode will be set to `AutoVsync` when Vsync is enabled, or `AutoNoVsync`
+    /// when Vsync is disabled. To set the present mode to `Mailbox` or another value, use the
+    /// [`Pixels::set_present_mode`] method.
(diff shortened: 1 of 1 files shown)


parasyte (approved):
Looks good.

Do you think you would ever want a getter for the present mode? It could save the caller from maintaining that state (since `Pixels` already stores it) and it's just a one-liner. All good if you don't think it's useful.

Kanabenki:
Good idea, added that!

parasyte (approved):
Thank you!

PR #375, opened 2023-10-01
Add pixel_to_window_pos() func

### LINKED ISSUE #366: Feature request: Getting window position from pixel position
I'm sure this usecase is very esoteric, but I'm using egui with pixels and I wanted to put egui objects on the same screen location as objects I'm drawing on the pixel buffer, but to do that I need some way of going from pixel positions to window positions, and surprisingly there doesn't seem to be a way right now, even though there is a way to go from window position to pixel position via `window_pos_to_pixel`. Even without a crazy usecase like that, there are probably other scenarios where this would be something you want.

omprakaash:
Added a function to obtain physical windows position of a pixel. Opening a draft PR to ensure correctness first. I am kind of rusty on graphical transformations atm. Should resolve #366 

 + Taking it out of draft.  Tested it by following these steps. Running example/minimal-winit -> obtaining physical_location of cursor when a cursor_move event occurs -> obtaining pixel_pos from that physical_location -> running the new `pixel_pos_to_window_pos() `func to see if the inital phycical_location of cursor matches with the window position obtained from the new function.

 It usually matches but is sometime off by one (possibly because a pixel could take up more than a single 'unit' of space in a window ? )

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,6 +37,7 @@ pub use raw_window_handle;
 use raw_window_handle::{HasRawDisplayHandle, HasRawWindowHandle};
 use thiserror::Error;
 pub use wgpu;
+use crate::renderers::ScalingMatrix;
 
 mod builder;
 mod renderers;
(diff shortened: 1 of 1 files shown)


parasyte:
This seems to share a lot of code with `window_pos_to_pixel()`. Do you see any way to DRY it up?

parasyte:
Closing this since it has diverged and there doesn't seem to be any interest.

PR #320, opened 2022-12-14
Bump wgpu from `0.13` to `0.14`

dtcristo:
Continuing on from #312 by @Zageron, I've addressed some feedback from that PR. I've also updates some crates in a few of the examples so they can now work with `wgpu 0.14`.

## Details

- Addressed feedback from #312.
- Fixed `imgui-winit` example by bumping a few crates and fixing some API changes.
- Fixed `minimal-tao` example by bumping `tao` from `0.12` to `0.15`.
- Fixed `minimal-fltk` example by bumping `fltk` from `1.3.13` to `1.3.17`.
- Fixed `minimal-egui` example by buumping `egui`, `egui-wgpu` and `egui-winit` from `0.19` to `0.20`. Also made some code changes to support this new version.
- Bumped MSRV from `1.61` to `1.65` as `egui 0.20` has dependencies that require `1.65`.
- Deleted `minimal-sdl2` example.
- Fixed new clippy lints.

PR #378, opened 2023-10-22
update wgpu to 0.17

barzamin:
the required changes are trivial. `wgpu::util::initialize_adapter_from_env`'s signature changed; the backend_bits parameter was removed. allowed backends are now fully constrained by the instance's enabled set (see [the changelog](https://github.com/gfx-rs/wgpu/blob/trunk/CHANGELOG.md#initialize_adapter_from_env-argument-changes)).


parasyte:
The new version of `egui` is requiring Rust 1.70. We can increment the MSRV for this. The last time this happened was in #320.

It's a bit challenging because the MSRV in the policy will need to be 1.70, but the crate itself can technically be on an older version (perhaps [remaining on 1.64](https://github.com/parasyte/pixels/blob/705f22b60d6bf720b4bd696f43e69f1ffa057921/Cargo.toml#L7)?) so that it can be used with older compilers when `egui` is not used.

barzamin:
I can bump `MSRV.md` and the CI config if you want, unless you'd prefer to touch the policy stuff yourself (I think the PR is editable by maintainers).

parasyte:
You are welcome to give it a shot if you'd like to see this land sooner than I could get around to it! I have some free time on Sunday which might be the soonest I could manage.

parasyte (approved):
Perfect! Thank you.

PR #380, opened 2023-10-29
Update `winit` to 0.28.7 in most examples (except `imgui-winit`)

nxsaken:
Updated most examples (related to https://github.com/rust-windowing/winit/issues/2876). The `imgui-winit` wasn't updated because it relies on `imgui-winit-support` which hasn't been updated in a while (see https://github.com/imgui-rs/imgui-rs/issues/749)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -497,7 +497,7 @@ dependencies = [
  "log",
  "pixels",
  "randomize",
- "winit 0.28.6",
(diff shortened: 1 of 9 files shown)


parasyte:
Hi! This doesn't actually do anything. It's enough to use `cargo update` to increase the locked version to the latest patch revision without changing any of the `Cargo.toml` files.

PR #381, opened 2023-10-29
Replace `line_drawing` with `clipline`

nxsaken:
Replaced `line_drawing` with my [`clipline`](https://github.com/nxsaken/clipline) crate which implements a variation of Bresenham's algorithm with built-in pixel-perfect clipping.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -394,6 +394,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "75476fe966a8af7c0ceae2a3e514afa87d4451741fcdfab8bfaa07ad301842ec"
 
 [[package]]
+name = "clipline"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "261dc8d360e39175d787ffc8ee5885da0ec7fe167171d574f6857772a348c344"
+
+[[package]]
 name = "cmake"
 version = "0.1.50"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


parasyte:
Does this crate offer better performance or something? It has no dependencies, which might be enough of an improvement to make this change worth it.

But I would strongly prefer an external iterator interface over the callback. I have written [a little about why](https://github.com/emilk/egui/issues/1790#issuecomment-1730667888) in another issue. There really isn't any reason to _not_ let the caller write the loop. And enforcing an internal iteration is a form of inversion of control which I am firmly against [on principle](https://youtu.be/PUv66718DII?si=421Q7iePXdOUe9g0).

If you switch it to external iteration, I don't see any reason not to make the change. Assuming it doesn't cause performance regressions or anything.

nxsaken:
Yes, this crate clips the line before the tight loop (in constant time), allowing you to skip the bounds checks when indexing into the pixel buffer. It also skips Bresenham's algorithm for vertical and horizontal lines. I haven't had experience writing non-trivial Iterators, so I avoided doing it for this crate - but I'll look into it. Thank you for sharing the discussion and the talk!

nxsaken:
I implemented the iterator(s) and updated the examples. You can either iterate over the `Clipline` enum itself, or match on it and move the loop(s) inside its arms, which will remove the overhead of the match from every iteration. I decided not to do that to keep the examples concise.

parasyte:
That looks very good! Thank you for making that change. I know it's a lot to ask, but I also believe it is a marked improvement for usability.

I'll let the CI run, and it should inform you if there is anything that needs to be addressed.

nxsaken:
Thank you for your feedback, that was valuable! It's my first crate, so I'm extra grateful for the opportunity to learn :)

parasyte (approved):
All looks good! This can be merged now.

PR #383, opened 2023-11-25
replace HasRaw*-Traits with Has* equivalents.

paulfrische:
I took the first steps to migrate to the newest version of raw-window-handle.
Have a great day!


parasyte:
Thank you! Please let me know when you are happy with this, and I will give it a full review. Please note that I don't think GH has ever sent me an email when PRs transition from Draft to Ready for Review. It will just sit around if you don't ping me!

PR #386, opened 2024-01-01
Update wgpu to 0.18

JacobMillward:
This allows the latest version of egui to be used. Required updating several `RenderPassDescriptor` objects to account for new fields, as well as updating to handle the new texture format `Rgb10a2Uint`.

Also updated the `minimal-egui` example.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -386,9 +386,9 @@ checksum = "75476fe966a8af7c0ceae2a3e514afa87d4451741fcdfab8bfaa07ad301842ec"
 
 [[package]]
 name = "clipline"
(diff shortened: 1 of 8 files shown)


parasyte:
@JacobMillward Heads up, there are CI failures that need to be fixed. It's the dependency chain for `imgui`. which needs https://github.com/Yatekii/imgui-wgpu-rs/pull/104 I would accept this with a git dependency reference for `imgui-wgpu` pointing at the git rev in that PR. I can't say it will be the last CI failure, though.

JacobMillward:
Superseded by #391

PR #390, opened 2024-01-25
Optimization: internally track buffer mutations

parasyte:
Adds an internal bool that tracks buffer mutations (with high probability). The pixel buffer is only written to the texture view when the flag has been set, and the flag is always cleared immediately before the upload begins.

Exchanges a potentially expensive memcpy with a comparatively free branch.

@dbalsom Here is the PR promised! Give it a try on your target device and let me know if this helps, or even if it makes things worse.

See: https://github.com/parasyte/pixels/discussions/387


parasyte:
Closing this, since there doesn't seem to be any interest.

Issue #338, opened 2023-02-24
imgui UI out of bounds in example "imgui-winit"


martinteoharov:
Whenever I switch between my monitors and run my program I get an instant crash before the first render. I narrowed it down to the UI drawing at a slightly bigger frame.

```
[2023-02-24T15:14:51Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    In a RenderPass
      note: encoder = `pixels_command_encoder`
    In a set_scissor_rect command
    Scissor Rect { x: 0, y: 0, w: 911, h: 650 } is not contained in the render target Extent3d { width: 910, height: 650, depth_or_array_layers: 1 }
```

parasyte:
It looks like an off by one based on the error message, but this is a new failure mode to me.

Do your displays have different logical pixel sizes? Maybe that has something to do with it. I honestly have not done a lot of testing with multiple displays and handling changes to the scaling factor. I can put together a setup and try to reproduce this.

Because that example uses the scaling factor within its resize handler, I suspect these two events need to be handled simultaneously. That's my current theory.

martinteoharov:
thanks @parasyte . My setup is basically a laptop that i connect to 2 different monitors (1920x1080 and 1920x1200) throughout the day (not at the same time).

ill have a look tomorrow and will try to come up with a solution, thanks for the advice and for the great project, ive found it very useful :))

tommythorn:
I just cloned the repo and ran through some examples and hit this.  My message was a bit different though:
```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 14.25s
     Running `/Users/tommy/projects/Rust/pixels/target/debug/imgui-winit`
[2024-11-12T22:16:56Z ERROR imgui_winit] pixels.resize_surface() failed: Texture width is invalid: 4294967295
[2024-11-12T22:16:56Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default
thread 'main' panicked at /Users/tommy/.cargo.x86_64-darwin/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.17.2/src/backend/direct.rs:3056:5:
wgpu error: Validation Error

Caused by:
    In a RenderPass
      note: encoder = `pixels_command_encoder`
    In a set_scissor_rect command
[...]

PR #391, opened 2024-01-31
Bump `wgpu` to 0.19 and `winit` to 0.29

mkrasnitski:
Also bumps `raw-window-handle` to 0.6. Blocked on the following:

- [x] A new `imgui` release
    - [ ] Adding support to `imgui-wgpu-rs`: https://github.com/Yatekii/imgui-wgpu-rs/pull/107
- [x] A new `egui` release
- [x] A new `fltk` release: https://github.com/fltk-rs/fltk-rs/issues/1512
- [x] Adding support to `game-loop`: https://github.com/tuzz/game-loop/pull/21

Right now this PR uses these libraries are used as git dependencies. I point at my own fork of [`imgui-wgpu-rs`](https://github.com/mkrasnitski/imgui-wgpu-rs) for now.

### DIFF
--- a/examples/minimal-tao/src/main.rs
+++ b/examples/minimal-tao/src/main.rs
@@ -4,11 +4,11 @@
 use error_iter::ErrorIter as _;
 use log::error;
 use pixels::{Error, Pixels, SurfaceTexture};
+use std::sync::Arc;
 use tao::dpi::LogicalSize;
 use tao::event::{Event, KeyEvent, WindowEvent};
 use tao::event_loop::{ControlFlow, EventLoop};
 use tao::keyboard::KeyCode;
-use tao::menu::{MenuBar, MenuItem};
 use tao::window::WindowBuilder;
 
 const WIDTH: u32 = 320;
@@ -27,25 +27,20 @@ fn main() -> Result<(), Error> {
     env_logger::init();
     let event_loop = EventLoop::new();
     let window = {
-        let mut file_menu = MenuBar::new();
-        file_menu.add_native_item(MenuItem::Quit);
-
-        let mut menu = MenuBar::new();
-        menu.add_submenu("File", true, file_menu);
-
         let size = LogicalSize::new(WIDTH as f64, HEIGHT as f64);
-        WindowBuilder::new()
+        let window = WindowBuilder::new()
             .with_title("Hello Pixels/Tao")
-            .with_menu(menu)
             .with_inner_size(size)
             .with_min_inner_size(size)
             .build(&event_loop)
-            .unwrap()
+            .unwrap();
+        Arc::new(window)
     };
 
     let mut pixels = {
         let window_size = window.inner_size();
-        let surface_texture = SurfaceTexture::new(window_size.width, window_size.height, &window);
+        let surface_texture =
+            SurfaceTexture::new(window_size.width, window_size.height, Arc::clone(&window));
         Pixels::new(WIDTH, HEIGHT, surface_texture)?
     };
     let mut world = World::new();
--- a/internals/pixels-mocks/src/lib.rs
+++ b/internals/pixels-mocks/src/lib.rs
@@ -1,57 +1,19 @@
 #![deny(clippy::all)]
 
-pub struct Rwh;
+pub struct Window;
 
-unsafe impl raw_window_handle::HasRawWindowHandle for Rwh {
-    fn raw_window_handle(&self) -> raw_window_handle::RawWindowHandle {
-        #[cfg(target_os = "macos")]
-        return raw_window_handle::RawWindowHandle::AppKit(
-            raw_window_handle::AppKitWindowHandle::empty(),
-        );
-        #[cfg(any(
-            target_os = "linux",
-            target_os = "dragonfly",
-            target_os = "freebsd",
-            target_os = "netbsd",
-            target_os = "openbsd",
-        ))]
-        return raw_window_handle::RawWindowHandle::Wayland(
-            raw_window_handle::WaylandWindowHandle::empty(),
-        );
-        #[cfg(target_os = "windows")]
-        return raw_window_handle::RawWindowHandle::Win32(
-            raw_window_handle::Win32WindowHandle::empty(),
-        );
-        #[cfg(target_os = "ios")]
-        return raw_window_handle::RawWindowHandle::UiKit(
-            raw_window_handle::UiKitWindowHandle::empty(),
-        );
+impl raw_window_handle::HasWindowHandle for Window {
+    fn window_handle(
+        &self,
+    ) -> Result<raw_window_handle::WindowHandle<'_>, raw_window_handle::HandleError> {
+        unimplemented!()
     }
 }
 
-unsafe impl raw_window_handle::HasRawDisplayHandle for Rwh {
-    fn raw_display_handle(&self) -> raw_window_handle::RawDisplayHandle {
-        #[cfg(target_os = "macos")]
-        return raw_window_handle::RawDisplayHandle::AppKit(
-            raw_window_handle::AppKitDisplayHandle::empty(),
-        );
-        #[cfg(any(
-            target_os = "linux",
-            target_os = "dragonfly",
-            target_os = "freebsd",
-            target_os = "netbsd",
-            target_os = "openbsd",
-        ))]
-        return raw_window_handle::RawDisplayHandle::Wayland(
-            raw_window_handle::WaylandDisplayHandle::empty(),
-        );
-        #[cfg(target_os = "windows")]
-        return raw_window_handle::RawDisplayHandle::Windows(
-            raw_window_handle::WindowsDisplayHandle::empty(),
-        );
-        #[cfg(target_os = "ios")]
-        return raw_window_handle::RawDisplayHandle::UiKit(
-            raw_window_handle::UiKitDisplayHandle::empty(),
-        );
+impl raw_window_handle::HasDisplayHandle for Window {
+    fn display_handle(
+        &self,
+    ) -> Result<raw_window_handle::DisplayHandle<'_>, raw_window_handle::HandleError> {
+        unimplemented!()
     }
 }
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,17 +1,16 @@
 use crate::renderers::{ScalingMatrix, ScalingRenderer};
 use crate::{Error, Pixels, PixelsContext, SurfaceSize, SurfaceTexture, TextureError};
-use raw_window_handle::{HasRawDisplayHandle, HasRawWindowHandle};
 
 /// A builder to help create customized pixel buffers.
-pub struct PixelsBuilder<'req, 'dev, 'win, W: HasRawWindowHandle + HasRawDisplayHandle> {
-    request_adapter_options: Option<wgpu::RequestAdapterOptions<'req>>,
+pub struct PixelsBuilder<'req, 'dev, 'win, W: wgpu::WindowHandle + 'win> {
+    request_adapter_options: Option<wgpu::RequestAdapterOptions<'req, 'win>>,
     device_descriptor: Option<wgpu::DeviceDescriptor<'dev>>,
     backend: wgpu::Backends,
     width: u32,
     height: u32,
     _pixel_aspect_ratio: f64,
     present_mode: wgpu::PresentMode,
-    surface_texture: SurfaceTexture<'win, W>,
+    surface_texture: SurfaceTexture<W>,
     texture_format: wgpu::TextureFormat,
     render_texture_format: Option<wgpu::TextureFormat>,
     surface_texture_format: Option<wgpu::TextureFormat>,
@@ -19,9 +18,7 @@ pub struct PixelsBuilder<'req, 'dev, 'win, W: HasRawWindowHandle + HasRawDisplay
     blend_state: wgpu::BlendState,
 }
 
-impl<'req, 'dev, 'win, W: HasRawWindowHandle + HasRawDisplayHandle>
-    PixelsBuilder<'req, 'dev, 'win, W>
-{
+impl<'req, 'dev, 'win, W: wgpu::WindowHandle + 'win> PixelsBuilder<'req, 'dev, 'win, W> {
     /// Create a builder that can be finalized into a [`Pixels`] pixel buffer.
     ///
     /// # Examples
@@ -30,7 +27,7 @@ impl<'req, 'dev, 'win, W: HasRawWindowHandle + HasRawDisplayHandle>
     /// use pixels::wgpu::{PowerPreference, RequestAdapterOptions};
     ///
     /// # use pixels::PixelsBuilder;
-    /// # let window = pixels_mocks::Rwh;
+    /// # let window = pixels_mocks::Window;
     /// # let surface_texture = pixels::SurfaceTexture::new(256, 240, &window);
     /// let mut pixels = PixelsBuilder::new(256, 240, surface_texture)
     ///     .request_adapter_options(RequestAdapterOptions {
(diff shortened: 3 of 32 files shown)


mkrasnitski:
There is one major bug that manifests in the `imgui-winit` example and seems to come from `imgui-wgpu-rs`. Resizing the window crashes with the following error:
```
[2024-01-31T01:42:15Z ERROR wgpu::backend::wgpu_core] Handling wgpu errors as fatal by default
thread 'main' panicked at /home/michael/.local/share/cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.19.1/src/backend/wgpu_core.rs:3009:5:
wgpu error: Validation Error

Caused by:
    In a RenderPass
      note: encoder = `pixels_command_encoder`
    In a set_scissor_rect command
    Scissor Rect { x: 0, y: 0, w: 641, h: 480 } is not contained in the render target (640, 480, 1)
```
Removing the `gui.handle_event` line from the example prevents the bug, so this is likely from `imgui-wgpu-rs`; also there's a call to `set_scissor_rect` internally in the lib.

Another annoyance that only happens with my fork is that key events are logged to the terminal like so:
```
KEY EVENT: KeyEvent { physical_key: Code(AltLeft), logical_key: Named(Alt), text: None, location: Left, state: Pressed, repeat: false, platform_specific: KeyEventExtra { key_without_modifiers: Named(Alt), text_with_all_modifiers: None } }
```
I don't know why this happens. I'd appreciate help diagnosing the cause of both of these problems, and would consider them a current blocker.

parasyte:
Hi, thanks for starting this! I'll kick off the CI pipeline shortly.

The difference between the scissor rect and surface has been a very common battle. Specifically I believe it's #338 (which still needs more info). The same issue was fixed in the `custom-shader` example in #286.

parasyte:
> Another annoyance that only happens with my fork is that key events are logged to the terminal like so:
> 
> ```
> KEY EVENT: KeyEvent { physical_key: Code(AltLeft), logical_key: Named(Alt), text: None, location: Left, state: Pressed, repeat: false, platform_specific: KeyEventExtra { key_without_modifiers: Named(Alt), text_with_all_modifiers: None } }
> ```

This is printed by the `imgui-winit-support` crate: https://github.com/imgui-rs/imgui-rs/blob/ca05418cb449dadaabf014487c5c965908dfcbdd/imgui-winit-support/src/lib.rs#L486

examples/minimal-tao/src/main.rs
```diff
@@ -27,25 +27,20 @@ fn main() -> Result<(), Error> {
...
-        file_menu.add_native_item(MenuItem::Quit);
-
-        let mut menu = MenuBar::new();
-        menu.add_submenu("File", true, file_menu);
-
         let size = LogicalSize::new(WIDTH as f64, HEIGHT as f64);
-        WindowBuilder::new()
+        let window = WindowBuilder::new()
             .with_title("Hello Pixels/Tao")
-            .with_menu(menu)
```
parasyte, line 39: It is sad to see this go, but I know exactly why it had to be: https://github.com/tauri-apps/tao/issues/717#issuecomment-1532055848

I would be inclined to follow up this PR by reintroducing the menu with the `muda` crate.

internals/pixels-mocks/src/lib.rs
```diff
@@ -1,57 +1,19 @@
...
-        );
-        #[cfg(target_os = "ios")]
-        return raw_window_handle::RawWindowHandle::UiKit(
-            raw_window_handle::UiKitWindowHandle::empty(),
-        );
+impl raw_window_handle::HasWindowHandle for Window {
+    fn window_handle(
+        &self,
+    ) -> Result<raw_window_handle::WindowHandle<'_>, raw_window_handle::HandleError> {
+        unimplemented!()
```
parasyte, line 9: Interesting! This should be ok because we don't want the tests to run against `pixels-mocks`, we just need to ensure they build. N.b. #80

src/builder.rs
```diff
@@ -243,14 +240,14 @@ impl<'req, 'dev, 'win, W: HasRawWindowHandle + HasRawDisplayHandle>
...
-    async fn build_impl(self) -> Result<Pixels, Error> {
+    async fn build_impl(self) -> Result<Pixels<'win>, Error> {
         let instance = wgpu::Instance::new(wgpu::InstanceDescriptor {
             backends: self.backend,
             ..Default::default()
         });
 
         // TODO: Use `options.pixel_aspect_ratio` to stretch the scaled texture
-        let surface = unsafe { instance.create_surface(self.surface_texture.window) }?;
+        let surface = instance.create_surface(self.surface_texture.window)?;
```
parasyte, line 250: This is ... incredible! Does that mean #238 is fixed, and we can add `#![forbid(unsafe_code)]` back to the lib?

src/builder.rs
```diff
@@ -625,5 +623,7 @@ const fn texture_format_size(texture_format: wgpu::TextureFormat) -> f32 {
 
         // 12x12 blocks, 16 bytes per block
         Astc { block: B12x12, channel: _ } => 9.0, // 12.0 * 12.0 / 16.0
+
+        NV12 => todo!(),
```
parasyte, line 627: This needs to be implemented. The size appears to be 3.

parasyte (commented):
Just some comments. I think the one `todo!` is worth addressing. But also, now that I review the value for other compressed texture formats, I'm not even sure if those are correct...

examples/minimal-tao/src/main.rs
mkrasnitski: I did try, but it seems like `muda` is a lot less cross-platform than `tao::menu` used to be. For instance, I tried replicating the old menu using `muda::PredefinedMenuItem::Quit`, but that type is unsupported on Linux.

parasyte: No worries, we can look at that separately!

internals/pixels-mocks/src/lib.rs
mkrasnitski: Part of the reason I did this is because all of the `::empty` methods on the various raw handle types got removed, so now you need to provide atleast a non-null pointer when constructing them. We could certainly try to make them runnable, though that might get involved.

mkrasnitski:
Looks like `egui 0.24` bumped MSRV to Rust 1.72. Since we're bounding types using the new versions of the rwh traits, we're not compatible with older versions of `egui` except the newest one (when it releases). So, I guess that means we should bump MSRV as well.

src/builder.rs
mkrasnitski: How do you figure that? I'm finding conflicting figures when trying to get the answer. Some say it's 12 bits per pixel - the wgpu documentation says that it consists of two 8-bit planes.

src/builder.rs
mkrasnitski: I'm surprised my approach worked. Since the handle traits have auto-impls for `&W`, I was able to switch from `&'win W` to `W + 'win`, which means you can pass owned objects which satisfy a `+ 'static` bound. This way I was able to overcome the borrow checker for some of the examples by using `Arc`.

mkrasnitski:
> This is printed by the `imgui-winit-support` crate: https://github.com/imgui-rs/imgui-rs/blob/ca05418cb449dadaabf014487c5c965908dfcbdd/imgui-winit-support/src/lib.rs#L486

That definitely seems like it was accidentally left in. Probably worth a PR to `imgui`.

src/builder.rs
parasyte: The WGPU docs say that it is two 8-bit planes, that's true, but the second plane is a _dual 8-bit channel_. It is 12-bits per pixel, there are just two pixels per "block". 24-bits total.

mkrasnitski: Ok, how should I comment the size amount then?

mkrasnitski:
Seems like CI is failing because of `impl_trait_projections`, which were stabilized in [Rust 1.74](https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#projections-in-opaque-return-types). I could change `Pixels::new_async` to explicitly name `Pixels<'win>` instead of `Self`, or we could bump MSRV again to 1.74.

src/builder.rs
parasyte: Side note: I could be totally wrong. But that's how I read it.

parasyte: So, I found this, which seems to be _the_ authoritative answer: https://github.com/gfx-rs/wgpu/blob/84ba4e5461fd1f305f8591cb086c40a488fbdb71/wgpu-hal/src/vulkan/conv.rs#L77

Which, if we lookup in Vulkan documentation at https://docs.vulkan.org/spec/latest/chapters/formats.html#formats-compatibility says:

> Block size 3 byte

And you can search for `VK_FORMAT_G8_B8R8_2PLANE_420_UNORM` to find the following prose description:

> `VK_FORMAT_G8_B8R8_2PLANE_420_UNORM` specifies an unsigned normalized multi-planar format that has an 8-bit G component in plane 0, and a two-component, 16-bit BR plane 1 consisting of an 8-bit B component in byte 0 and an 8-bit R component in byte 1. The horizontal and vertical dimensions of the BR plane are halved relative to the image dimensions, and each R and B value is shared with the G components for which \(\left\lfloor i_G \times 0.5 \right\rfloor = i_B = i_R\) and \(\left\lfloor j_G \times 0.5 \right\rfloor = j_B = j_R\). The location of each plane when this image is in linear layout can be determined via [vkGetImageSubresourceLayout](https://docs.vulkan.org/spec/latest/chapters/resources.html#vkGetImageSubresourceLayout), using `VK_IMAGE_ASPECT_PLANE_0_BIT` for the G plane, and `VK_IMAGE_ASPECT_PLANE_1_BIT` for the BR plane. This format only supports images with a width and height that is a multiple of two.

mkrasnitski: Alright, I'll put it down as a 24-bit format then.

parasyte:
```suggestion
        // 8-bit two-plane 4:2:0 YUV
        NV12 => 3.0, // (8.0 + 16.0) / 8.0
```

parasyte:
```suggestion
        // 8-bit two-plane 4:2:0 YUV
        // The first plane consists of 8-bit G components. the second plane consists of 16-bit BR components. The resolution of the second plane is halved both vertically and horizontally.
        NV12 => 1.5, // (8.0 + 16.0 / 2.0 / 2.0) / 8.0
```

parasyte: Hi, sorry, could you update this again? I worked it out in my head, and your original finding of 12-bits per pixel is correct. It isn't 24-bits.

mkrasnitski: Done.

mkrasnitski:
Oh, didn't see the other "MSRV"-style checks. I guess I should bump the 1.70 test version to 1.74, but the Rust-version check seems redundant.

parasyte:
That check is there to handle this clause from our MSRV policy:

> The MSRV will be chosen as the minimum version of `rustc` that can successfully pass CI, including documentation, lints, and all examples. For this reason, the minimum version _supported_ may be higher than the minimum version _required_ to compile the pixels crate itself.

Because dependencies in the examples tend to require a newer version of the compiler than the library needs, testing the two versions independently allows the library to require a compiler version that diverges from the examples.

Using the git HEAD is an unfortunate case of laziness where I just never added a way to select the correct branch or PR.

There is more information on this setup here:

- #295 
- https://github.com/parasyte/pixels-ci-rust-version

So, no. It is definitely not redundant.

parasyte:
It is valid criticism and thank you for the reminder. I am currently OK with doing the maintainer thing and keeping CI happy post-merge with additional PRs.

lukexor:
Excitingly looking forward to this merging as I'm working on a project to add egui to my existing winit project using pixels. However, I'm concerned about the new lifetime requirement on `Pixels` as that has a very wide ranging impact on any application with a lot of impl blocks for any struct containing the pixels instance.

mkrasnitski:
Yes, I agree that docs the should make it more obvious that to avoid lifetime headaches you should just wrap your window in an `Arc` and use `Pixels<'static>`. Suggestions welcome.

JacobMillward:
This seems to supersede my PR at #386 . I'm not quite sure what we're waiting on for this PR now, can someone explain?

mkrasnitski:
This PR is currently blocked on `imgui-wgpu-rs` supporting wgpu 0.19 and winit 0.29, so that the `imgui-winit` example compiles. Which should be easy, however it requires that lib changing to take a git dependency on `imgui-rs`, which got support `wgpu 0.19` a few months ago but hasn't seen a release since April 2023. I could open a PR to see if they're OK with that.

parasyte:
I would be in support of any effort to upstream some of the unblocking work. Speaking of, I should be able to make some more progress on https://github.com/imgui-rs/imgui-rs/pull/716 now!

vincenthz:
Would it make sense to drop imgui example (temporarily) and not block this PR ?

Once imgui catch up, the example can be added back, in the meantime imgui user can use the existing old version ?

parasyte:
I believe that is probably the right compromise. Both `wgpu` and `winit` have released new versions while this has been in limbo.

My proposal would be updating the imgui example to use the previous release version, with a comment somewhere about the transitive dependency issue.

mkrasnitski:
As if on cue, `imgui` had its 0.12 release 2 days ago. All that remains now is https://github.com/Yatekii/imgui-wgpu-rs/pull/107.

mkrasnitski:
From what I can tell, supporting `winit 0.30` in `pixels` would require similar efforts for adding support to other dependencies. The ecosystem hasn't yet moved over.

parasyte:
Yeah, upgrading `winit` is going to be kind of rough.

They plan to remove the existing event hander callback and replace it with a bucket of callbacks defined in a trait: https://github.com/rust-windowing/winit/issues/3626. Somehow digging deeper into Inversion of Control territory is appealing to the maintainers.

I'm keen to stick with 0.29 for now, but I'll be planning on a general replacement since they insist on making app and game development more difficult than it should be.

jeapostrophe:
Is there a way to use this with only winit and game-loop? I don't care about imgui, egui, or fltk. It is frustrating to wait so long to update when the release is held up by features that don't affect me.

mkrasnitski:
> Is there a way to use this with only winit and game-loop? I don't care about imgui, egui, or fltk. It is frustrating to wait so long to update when the release is held up by features that don't affect me.

Certainly - just point at my fork. Compatibility with dependencies you aren't using won't matter for you so you'll be able to use the lib fine.

jeapostrophe:
@mkrasnitski Thank you!

JacobMillward:
> Is there a way to use this with only winit and game-loop? I don't care about imgui, egui, or fltk. It is frustrating to wait so long to update when the release is held up by features that don't affect me.

I tend to agree with that there is an underlying issue here. The core library can be upgraded, it is merely the surrounding example integrations that consistently keep this project out of date. 

@parasyte it might be worth considering moving complex examples to pinned versions of the library, or moving them into a separate repository?

mkrasnitski:
Ok I'm fed up waiting for imgui-wgpu-rs to update support for wgpu, so the solution is to just pin the example to use a specific `pixels` commit instead of using `path`, and keep the dependencies on the old versions. This PR has been sitting blocked on transitive dependencies for far too long.

Btw @parasyte it might be worth it to cut release 0.13.1 off the current project HEAD and publish it so that the imgui example can point to `0.13.1` instead of a specific git commit.

mkrasnitski:
Sorry for the ping @parasyte, do you have any additional concerns, or is this PR free to merge?

parasyte:
My apologies for dropping the ball on this! I've had my attention focused elsewhere.

Just about to publish a new version from `main`. It required some version bumps and will lead to some conflicts.

parasyte:
I recognize how much effort this was, and how important the updates are. All of the checks are green, so I'm comfortable merging without hesitation. Any other needs can be addressed in smaller PRs.

Thank you for everything you did here, @mkrasnitski!

Issue #270, opened 2022-04-08
game_loop and winit_input_helper do not appear to be compatible (invaders example)


Zageron:
I was using your invaders example as a reference and ran into an issue.
This isn't something you can fix, but thought I'd let you know.

`g.game.input.update(&event)` never returns true, which means the state can never be guaranteed to be cleared.

Ex. This will never run the keyboard events, and if I remove the if guard the keyboard state remains between events and I get multiple false positives.

The only state you care about is the example is key_pressed and window_resized, for actions that don't care about repeat triggers. Add a print statement to see what I mean.

```rust
game_loop(
    event_loop,
    window,
    game,
    FPS,
    0.1,
    |g| {
        g.game.update();
    },
    |g| {
        // Drawing
        g.game.world.draw(g.game.pixels.get_frame());
        if let Err(e) = g.game.pixels.render() {
            error!("pixels.render() failed: {}", e);
            g.exit();
        }

        #[cfg(not(target_arch = "wasm32"))]
        {
            // Sleep the main thread to limit drawing to the fixed time step.
            // See: https://github.com/parasyte/pixels/issues/174
            let dt = TIME_STEP.as_secs_f64() - Time::now().sub(&g.current_instant());
            if dt > 0.0 {
                std::thread::sleep(Duration::from_secs_f64(dt));
            }
        }
    },
    |g, event| {
        if g.game.input.update(&event) {
            println!("{:?}", event);
            // Close events
            if g.game.input.key_pressed(VirtualKeyCode::Escape) || g.game.input.quit() {
                g.exit();
            }

            if g.game.input.key_released(VirtualKeyCode::Space) {
                g.game.head.play_song();
                println!("Playing song");
            }

            // Resize the window
            if let Some(size) = g.game.input.window_resized() {
                g.game.pixels.resize_surface(size.width, size.height);
            }
        }
    },
);
```

parasyte:
Looks like doing your own input accumulation is better in the short term, anyway. The current behavior of `winit_input_helper`'s `key_pressed` method isn't as documented: https://github.com/rukai/winit_input_helper/issues/17

Zageron:
Wow thanks for the insanely thorough investigation and quick fix attempt. :o

Zageron:
I tested our your branch as a dependency and it resolved my issues, thank you! :)
Once I get more progress done on my MVP I'll likely take into consideration your recommendation to refactor out `winit_input_helper` and write my own control logic.

[...]

PR #400, opened 2024-09-23
Bump dependencies for imgui-winit example

mkrasnitski:
[imgui-wgpu-rs](https://github.com/Yatekii/imgui-wgpu-rs) now supports wgpu 0.19 and winit 0.29, but only on the git HEAD. A release hasn't been cut yet so keeping this as draft for now. Also, resizing the window now crashes the app with the following error:

```
wgpu error: Validation Error

Caused by:
    In a RenderPass
      note: encoder = `pixels_command_encoder`
    In a set_scissor_rect command
    Scissor Rect { x: 0, y: 0, w: 642, h: 483 } is not contained in the render target (640, 480, 1)
```

I figured this PR would be a good place to track this while the root cause is figured out.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,15 +19,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c71b1793ee61086797f5c80b6efa2b8ffa6d5dd703f118545808a7f2e27f7046"
 
 [[package]]
-name = "addr2line"
-version = "0.24.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f5fb1d8e4442bd405fdfd1dacb42792696b0cf9cb15882e5d097b742a676d375"
-dependencies = [
- "gimli 0.31.0",
-]
-
-[[package]]
 name = "adler"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


parasyte:
The validation error is from desynchronized draw and resize. Resize is deferred by `winit-input-helper`, but drawing is not.

It can be fixed by reordering these operations. E.g. by moving the drawing code after the resize, but before the `world.update()` call. Or alternatively and preferably, move the resize out of the `input.update()` guard and use the normal `winit` event matching instead.

*Sigh.* These are some skeletons in the closet. The history here is sort of documented in https://github.com/parasyte/pixels/issues/270#issuecomment-1092627678.

I ended up matching on the event for `RedrawRequested` separately from using `input.update()`'s return value for some of the reasons described there. We kind of got away with it, but these events should not be deferred.

Deferring these events causes problems like the resize looking _horrible_ while the user drags the window frame around. The intermediate frame gets stretched and warped. The right way to handle resize will correctly redraw every frame while the resize is in progress. And doing that means no deferral of resize and redraw events.

And yeah, deferring events is what you are expected to do with `winit-input-helper`: https://github.com/rukai/winit_input_helper/blob/106e033c718ccfb346866c054d23025666ae13ae/examples/example.rs#L33-L35

So, it's probably time for it to go.

parasyte:
I'm also completely misremembering how it ended up that `RedrawRequested` is treated differently from the rest of the event handling. Because this was there since the beginning with a salient comment:

https://github.com/parasyte/pixels/blob/17c6054f40daf0a77544994c4d0e46ae4fe258d0/examples/invaders/main.rs#L34-L38

All of the examples inherited it.

parasyte:
Superseded by #433

PR #402, opened 2024-10-08
feat: upgrade wgpu to 22.0

clearlysid:
`wgpu` versions now use the `x.0` rather than the `0.x` scheme ie.. `22.0` is the new `0.22`.

Unfortunately, Cargo treats these as very different dependencies and versions. This causes the `pixels` crate to become incompatible with projects that are using a newer version of `wgpu` for something else.

In this PR, I've upgraded the wgpu version to latest and fixed the relevant API calls with newer args that the compiler mandated (just defaults for now).

Feel free to merge this if you feel this is relevant. Thanks for the great project! 🙌

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,6 +161,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ash"
+version = "0.38.0+1.3.281"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0bb44936d800fea8f016d7f2311c6a4f97aebd5dc86f09906139ec848cf3a46f"
+dependencies = [
+ "libloading 0.8.5",
+]
+
+[[package]]
 name = "async-broadcast"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


parasyte:
Yeah, `wgpu` still needs to be updated. This is a pretty good start, thank you!

I'm holding off any other updates to `wgpu` until the next release (`pixels` 0.15) with support for `wgpu` 0.19. Which still needs some work as discussed in #400.

But I'm glad this has already been started!

clearlysid:
Makes total sense!

mkrasnitski:
One complication with this is the `egui` example. Support for wgpu v22 was added in egui 0.29, which also bumps winit to 0.30. So, that example would need to be reworked to use the newer winit version. Also, seems like `winit_input_helper` doesn't yet support winit 0.30. There are some open PRs but nothing has been merged yet.

parasyte:
See also: #411 and #403

djsell:
I'm curious what the path forward here is, and how I may help. wgpu is now up to 24.0 (they're breaking things fast 😓 )

parasyte:
@djsell What's driving the need to upgrade `wgpu` on your end? Are you trying to gain access to new underlying features or bug fixes? Does your use case integrate with other crates that you've had to upgrade for other reasons? I'm trying to get a read on the motivation and urgency.

djsell:
There may have been a bug fix in there as well that I needed to upgrade for, but I have forgotten at this point. I've been using sid's branch for a while.

parasyte:
As of now, [`iced_wgpu` still uses `wgpu = "0.19"`](https://crates.io/crates/iced_wgpu/0.13.5/dependencies) and [`vello` is at `wgpu = "23"`](https://crates.io/crates/vello/0.4.0/dependencies). Finding the intersection between transitive dependencies is pretty painful.

Ideally, we should have at least one release for each `wgpu` version. 🤔 I was able to keep up while I was still figuring out the APIs. But now that it's basically stable, these releases are just a maintenance cost.

nicoburns:
I'm tracking `vello` and also `bevy`s WGPU versions (luckily Vello is tracking Bevy's version for compatibility, so these tend to be the same). These are currently both at `24` and are likely to to be the newly released `26` in the next release. Another motivation for this for me is to keep the version of things like the `windows` crate up-to-date (to avoid duplicate version in my crate graph).

PR #410, opened 2025-01-02
Add basic nix flake

Juanperias:
this pull request adds a fairly simple nix flake that meets the needs of having the rust toolchain installed, plus it adds a rust-toolchain.toml file that includes the wasm and android targets.

you can use the flake with:
```
nix develop
```

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,130 @@
+{
+  "nodes": {
+    "crane": {
+      "locked": {
+        "lastModified": 1734808813,
+        "narHash": "sha256-3aH/0Y6ajIlfy7j52FGZ+s4icVX0oHhqBzRdlOeztqg=",
+        "owner": "ipetkov",
+        "repo": "crane",
+        "rev": "72e2d02dbac80c8c86bf6bf3e785536acf8ee926",
+        "type": "github"
+      },
+      "original": {
+        "owner": "ipetkov",
+        "repo": "crane",
+        "type": "github"
+      }
+    },
+    "fenix": {
+      "inputs": {
+        "nixpkgs": "nixpkgs",
+        "rust-analyzer-src": "rust-analyzer-src"
+      },
+      "locked": {
+        "lastModified": 1735799625,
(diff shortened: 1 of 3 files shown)


parasyte:
Hi. The concern I have for this is that the toolchain file locks the project to the stable release channel. Our CI is setup to build and test [with multiple channels](https://github.com/parasyte/pixels/blob/d4df286d3f886e9b12fb36f731555c3be66c8ebd/.github/workflows/ci.yml#L14-L16). The existence of the toolchain file will override the toolchains used in the CI jobs.

And I don't know if there is a way to "override the override". Let alone getting into situations where you have to override the override for the override. It's best to just not force an override in the first place.

----

Other than that, I don't use nix. I cannot fully appreciate any benefits that this PR is adding. `pixels` is just a library with infrequent releases. The dependents are going to lock the specific version they are using. It's unclear to me how nix fits in, or why a secondary package manager is needed.

I guess I'm indifferent to the nix configuration. It's kind of like committing a VS Code project file or something. Just unused files in the repo that make the file list longer.

Juanperias:
Hi, thanks for commenting, I have made some changes to avoid some problems, in this case and the most important is to remove the rust-toolchain to avoid problems with the CI.

I understand your point of not understanding the need to use Nix, in my case it was because I wanted to contribute to this project, and I created this simple flake to have the rust toolchain, so I think it could be useful for other Nix users who want to contribute to this project.

parasyte:
Ok, sorry for taking so long to get back to this.

My follow up question is how will this work out for maintaining the flake and lock file? I have no intention to do so. Will it become outdated or unusable over time if it isn't kept in sync with upstreams?

I'm ok accepting this, but I want to know if you or someone else is interested in taking up the maintenance burden.

Juanperias:
Yes, I would be willing to keep the flake updated. To answer your question, it is possible that the flake could become obsolete over time if it isn't kept in sync with the upstreams.

(By the way, sorry for answering three weeks later)

parasyte (approved):

parasyte:
I'm going to close this. I don't have a need for it and no one has shown any interest. Feel free to reopen. But I will still need someone else to maintain it.

Issue #423, opened 2025-06-01
Cannot get SurfaceCapabilities from pixels


alago1:
When I try to run my code on WSL with PresentMode::Immediate, I get the following error:
```
thread 'main' panicked at .cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-0.19.4/src/backend/wgpu_core.rs:724:18:
Error in Surface::configure: Validation Error

Caused by:
    Requested present mode Immediate is not in the list of supported present modes: [Fifo]

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
This is fine and expected but I often run this code on both the WSL environment that is capped to 60Hz and in my Windows environment that can actually run with PresentMode::Immeditate.

I want to conditionally set PresentMode::Immediate when it's available otherwise default to Fifo. wgpu supports getting this information under the `SurfaceCapabilities` struct ([docs](https://docs.rs/wgpu/latest/wgpu/struct.SurfaceCapabilities.html)) but this is called on the Surface instance that is not exposed by pixels as far as I've checked.

Would it be possible to expose this information through Pixels somehow? One option would be to add a reference to the surfaceCapabilties to the PixelsContext struct. It looks like the SurfaceCapabilties is already used to build the PixelsContext+Pixels instance: https://github.com/parasyte/pixels/blob/main/src/builder.rs#L285

alago1:
thanks for reviewing the PR!

PR #424, opened 2025-06-01
add SurfaceCapabilities to PixelsContext

alago1:
Small change to expose the SurfaceCapabilities struct used when building the Pixels instance through the PixelsContext.

This should allow conditionally setting wgpu PresentMode based on the available options for the Pixels surface (Issue #423).

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -328,6 +328,7 @@ impl<'req, 'dev, 'win, W: wgpu::WindowHandle + 'win> PixelsBuilder<'req, 'dev, '
             texture_format: self.texture_format,
             texture_format_size: texture_format_size(self.texture_format),
             scaling_renderer,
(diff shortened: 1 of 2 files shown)


parasyte (approved):
Thank you for the contribution!

parasyte:
It looks like this will technically allow the developer to "fix up" `PresentMode` after the surface has already been configured.

Another option is to do the obvious thing when creating the surface: If the chosen `PresentMode` is not already in the list of capabilities, replace it with the first one available. Any other complex fallback logic can be delegated to user code with the new `PixelsContext::surface_capabilities` field. WDYT?

parasyte:
Sorry for the delay. I’m returning home from vacation right now. The change looks good and I’ll get this merged as I find time.

PR #433, opened 2026-01-19
Upgrade to wgpu 27 and ultraviolet 0.10

nicoburns:
This WGPU version is currently widely used in the ecosystem, including by Iced, Egui, Bevy, and Vello, and Blitz.

I'm not sure what to do about examples. I already updated the ones using WGPU. I would suggest updating all of the examples, and then immediately publishing a new version of pixels.


parasyte:
> Is there a reason for wanting such a conservative MSRV? I know there are domains where this is really important, but most of the games/UI ecosystem I'm familiar with seems to be quick to update rustc versions.

Some people are still managing their Rust installations through distro package managers. :(

> I'll see if I can fix up the rest of the examples.

Sounds good. I have been looking forward to pinning the examples to a published version of `pixels` to allow them to diverge.

nicoburns:
@clearlysid Any chance of a crates.io release with this update?

PR #451, opened 2026-04-29
Drop surface before reconfiguring it if texture is Suboptimal

### LINKED ISSUE #450: Getting wgpu ValidationError panic on Linux
I have a little project I've just updated to `pixels` 0.17 and while it works fine on MacOS, I get a panic on Linux like the following:

```
thread 'main' (32262) panicked at /home/amx/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-29.0.1/src/backend/wgpu_core.rs:3879:18:
wgpu error: Validation Error

Caused by:
  In Surface::configure
    `SurfaceOutput` must be dropped before a new `Surface` is made


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

After a bit of searching I found [this issue](https://github.com/gfx-rs/wgpu/issues/9360), so a similar fix should be made.

abusch:
Apply the same fix as https://github.com/gfx-rs/wgpu/pull/9384

closes: #450

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -556,9 +556,14 @@ impl<'win> Pixels<'win> {
             match self.context.surface.get_current_texture() {
                 CurrentSurfaceTexture::Success(surface_texture) => break surface_texture,
 
-                CurrentSurfaceTexture::Suboptimal(_)
-                | CurrentSurfaceTexture::Outdated
-                | CurrentSurfaceTexture::Lost => {
+                CurrentSurfaceTexture::Suboptimal(surface) => {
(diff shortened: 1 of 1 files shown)


parasyte:
Looks good. Can you add a comment explaining the purpose of the `drop`? Since it's critical and there doesn't appear to be a way to test this, I don't want it to accidentally regress.

abusch:
Done.

parasyte:
Thank you! 0.17.1 was just published with this fix: https://crates.io/crates/pixels/0.17.1

PR #457, opened 2026-07-11
Fix window-to-pixel coordinates when downscaling

skrmc:
Always use the texture center, fixes #456.


parasyte:
Thanks!

parasyte:
Just published a new release today containing this patch:

- https://github.com/parasyte/pixels/releases/tag/0.17.2
- https://crates.io/crates/pixels/0.17.2

